PR #12, opened 2019-10-14
Add pytest as install dependency

petemoore:
I had problems running tc-admin, which this change fixed.

I'm not sure why `pytest` is needed when just running tc-admin, but it seemed to help. This could be the wrong fix, so please check I put the dependency in the right place. Thanks!

This is the script I was running:

```bash
#!/bin/bash

cd "$(dirname "${0}")"

rm -rf tc-admin
mkdir tc-admin
pip3 install --upgrade pip

cd tc-admin
python3 -m venv tc-admin-venv
source tc-admin-venv/bin/activate
pip3 install --upgrade pip

git clone https://github.com/mozilla/community-tc-config
cd community-tc-config
pip3 install -e .
which tc-admin
TASKCLUSTER_ROOT_URL=https://community-tc.services.mozilla.com tc-admin diff
```

This is the output I was getting:

```
Requirement already up-to-date: pip in /usr/local/lib/python3.7/site-packages (19.2.3)
Collecting pip
  Using cached https://files.pythonhosted.org/packages/30/db/9e38760b32e3e7f40cce46dd5fb107b8c73840df38f0046d8e6514e675a1/pip-19.2.3-py2.py3-none-any.whl
Installing collected packages: pip
  Found existing installation: pip 19.0.3
    Uninstalling pip-19.0.3:
      Successfully uninstalled pip-19.0.3
Successfully installed pip-19.2.3
Cloning into 'community-tc-config'...
remote: Enumerating objects: 89, done.
remote: Counting objects: 100% (89/89), done.
remote: Compressing objects: 100% (60/60), done.
remote: Total 89 (delta 49), reused 50 (delta 24), pack-reused 0
Unpacking objects: 100% (89/89), done.
ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'requirements.txt'
Obtaining file:///Users/pmoore/tc-admin/community-tc-config
Collecting tc-admin~=1.0.0 (from community-tc-config==1.0.0)
  Using cached https://files.pythonhosted.org/packages/dc/ce/97b0212191ac6c675b62b9e92e9d405d64e6f7514ec3de287def0e92eba9/tc_admin-1.0.0-py3-none-any.whl
Collecting json-e~=3.0.0 (from community-tc-config==1.0.0)
  Using cached https://files.pythonhosted.org/packages/98/37/efe9813b85989e716f8cde7ab716f17ff98e1e18f68666d522eadd4680a8/json-e-3.0.0.tar.gz
[...]

PR #14, opened 2019-10-14
update tc-admin

djmitche:


djmitche:
Pete, can you check this fixes the issue you saw in #12?

djmitche:
https://github.com/taskcluster/tc-admin/releases/tag/v1.1.0

milescrabill:
I tested w/ both `1.0.0` and `1.1.0` replicating Pete's config from slack messages and interestingly, both worked for me.

djmitche:
Yeah, it's dependent on circumstances, which is why I missed it to start with.  I think if you run `python setup.py test` then it installs pytest, so if you do that before checking for this bug, you won't see it.  I think :)

Thanks!

PR #19, opened 2019-10-16
Bug 1588873 - load config from YAML files

djmitche:
This uses a variant of the `grants.yml` from ci-configuration to grant scopes to all sorts of roles.

And, moves projects to `projects.yml`, including their own grants.

And, updates worker generation to use some Python functions keyed by a `type` field in `projects.yml`.

As written, this wants to delete a bunch of in-progress roles for servo, deepspeech, webrender, etc.  Hopefully those will get ironed out, but in the interim be careful applying this.


config/projects.yml
```diff
@@ -0,0 +1,32 @@
...
+#     - role1
+#     - ..
+#   worker-pools:
+#     <worker-pool-name>:
+#       owner: ..
+#       emailOnError: ..
+#       type: <name of a function in `generate/workers.py`)
+#       ..: ..  # arguments to that function
+#
+# The worker-pool configurations are defined by Python functions, keyed by the `type` property.
```
imbstack: I like this solution

config/grants.yml
```diff
@@ -0,0 +1,86 @@
...
+    - auth:delete-role:hook-id:project-<..>/*
+    - auth:update-role:hook-id:project-<..>/*
+    - hooks:modify-hook:project-<..>/*
+    - hooks:trigger-hook:project-<..>/*
+
+    # project-specific index routes
+    - index:insert-task:project.<..>.*
+    - queue:route:index.project.<..>.*
+
+    # project-specific worker pools and workers
```
imbstack: These still need the * from the other PR right

generate/workers.py
```diff
@@ -0,0 +1,96 @@
...
+
+GCP_PROVIDER = 'community-tc-workers-google'
+DEFAUlT_GCP_DOCKER_WORKER_IMAGE = "projects/taskcluster-imaging/global/images/docker-worker-gcp-googlecompute-2019-10-08t02-31-36z"
+
+GOOGLE_REGIONS_ZONES = {
+    "us-east1": ["b", "c", "d"],
+    "us-east4": ["a", "b", "c"],
+}
+
+GOOGLE_ZONES_REGIONS = [
```
imbstack: Haha, I had a hunch this would become a comprehension when you fixed it

generate/workers.py
```diff
@@ -0,0 +1,96 @@
...
+        owner=cfg.get('owner', 'nobody@mozilla.com'),
+        emailOnError=cfg.get('emailOnError', False),
+        **wp)
+
+
+def base_google_config(**cfg):
+    return {
+        'providerId': GCP_PROVIDER,
+        'config': {
+            "maxCapacity": 20,
```
imbstack: Should we avoid having these default to something and instead make them always be specified as arguments?

djmitche: Oh, good point.  I'll do a followup.

PR #23, opened 2019-10-22
Add docker-worker scopes to run CI tests

walac:


config/projects.yml
```diff
@@ -55,12 +55,49 @@ taskcluster:
...
         - docker-worker:capability:device:loopbackAudio
         - docker-worker:capability:device:loopbackVideo
-        - project/taskcluster/testing/docker-worker/ci-creds
-        - project/taskcluster/testing/docker-worker/pulse-creds
+        - docker-worker:image:localhost:*
+        - secrets:get:project/taskcluster/testing/docker-worker/ci-creds
+        - secrets:get:project/taskcluster/testing/docker-worker/pulse-creds
+        - auth:sentry:test-dummy-worker
+        - auth:statsum:test-dummy-worker
+        - auth:webhooktunnel
```
djmitche: This scope doesn't actually do anything anymore -- the websocktunnel token endpoint's scope now has an audience and wstClientId in it.

config/projects.yml
```diff
@@ -55,12 +55,49 @@ taskcluster:
...
+        - queue:claim-work:test-dummy-provisioner/dummy-worker-*
+        - queue:create-task:test-dummy-provisioner/dummy-worker-*
+        - queue:create-task:null-provisioner/*
+        - queue:define-task:test-dummy-provisioner/dummy-worker-*
```
djmitche: Would it be very difficult to put these under a worker-pool name like `proj-taskcluster/dummy-*`?  Similar for `no-provisioning-nope`

config/projects.yml
```diff
@@ -55,12 +55,49 @@ taskcluster:
...
+        - queue:poll-task-urls:test-dummy-provisioner/dummy-worker-*
+        - queue:schedule-task:test-dummy-scheduler/*
+        - queue:task-group-id:test-dummy-scheduler/*
+        - queue:worker-id:test-dummy-workers/dummy-worker-*
+        - secrets:get:project/taskcluster/taskcluster-worker/stateless-dns
+        - worker:cache:dummy-cache-*
+        - queue:create-task:proj-taskcluster/ci
+        - queue:worker-id:random-local-worker/dummy-worker-*
+        - queue:claim-task:null-provisioner/*
+        - queue:worker-id:docker-worker/docker-worker
```
djmitche: Similar thing here, with workerGroup `proj-taskcluster`

config/projects.yml
```diff
@@ -55,12 +55,49 @@ taskcluster:
...
+        - queue:create-task:test-dummy-provisioner/dummy-worker-*
+        - queue:create-task:null-provisioner/*
+        - queue:define-task:test-dummy-provisioner/dummy-worker-*
+        - queue:get-artifact:SampleArtifacts/_/X.txt
+        - queue:poll-task-urls:test-dummy-provisioner/dummy-worker-*
+        - queue:schedule-task:test-dummy-scheduler/*
+        - queue:task-group-id:test-dummy-scheduler/*
+        - queue:worker-id:test-dummy-workers/dummy-worker-*
+        - secrets:get:project/taskcluster/taskcluster-worker/stateless-dns
+        - worker:cache:dummy-cache-*
```
djmitche: This is also an unusual scope -- does docker-worker pay attention to this?

config/projects.yml
walac: Can we do that in a followup patch? I would like to get things working first as is first.

djmitche: Sounds good -- can you make a bug or issue for it?

walac: ok

PR #25, opened 2019-10-22
add mozilla project

milescrabill:
this is pretty broad, so it's up for debate if this is the right approach.

this essentially allows any project in the mozilla organization access to this worker-pool

### DIFF
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -115,7 +115,7 @@ mozilla:
   adminRoles: []
   workerPools:
     ci:
-      owner:
+      owner: taskcluster-notifications+workers@mozilla.com
       emailOnError: false
       type: standard_gcp_docker_worker
       minCapacity: 1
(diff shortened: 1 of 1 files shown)


milescrabill:
I agree, that sounds like a good criteria for when a project should split off. I've renamed the project "mozilla-misc" which is still broad, but captures some of the point - people with access to the Mozilla org should be able to easily prototype using Taskcluster . If they need any more complexity, they should get their own project added.

config/projects.yml
```diff
@@ -75,3 +75,18 @@ bors-ng:
   grants:
     - grant: queue:create-task:highest:proj-bors-ng/ci
       to: repo:github.com/bors-ng/bors-ng:*
+
+# catch-all repo for simple mozilla projects
+# if you need more complexity, talk to the taskcluster team!
+mozilla-misc:
+  adminRoles: []
```
djmitche: Let's add `github-team:taskcluster/core` here, so that we can admin this project.

milescrabill: I tested doing this and got `RuntimeError: duplicate resources: Role=github-team:taskcluster/core` running `tc-admin diff`, I saw you added merging in `tc-admin` `v1.2.0` so I bumped our version to that.

djmitche: I just landed a PR to upgrade to 2.0.0 so that will help too.  Thanks :)

PR #28, opened 2019-10-23
Bug 1574664 - Add redo project

helfi92:


config/projects.yml
```diff
@@ -1,17 +1,17 @@
...
 # Each entry in this file is defined as follows:
 #
 # <project-name>:
-#   admin-roles:
+#   adminRoles:
 #     # roles with administrative access to the project's resources
 #     - role1
 #     - ..
-#   worker-pools:
+#   workerPools:
```
djmitche: Thanks!

helfi92:
> This looks good! We might also want to put this in the misc project that miles just added. We could do that now, or can follow up later.

As in duplicate the entry? What's the reason? I'm sure I'm missing something.

djmitche:
We can chat more this afternoon, but the idea would be to add scopes for `redo` using the `misc` project's resources, so for example it would use `proj-misc/ci` as its workerpool.  The `misc` project is sort of a (newly-invented) catch-all for small things like `redo` which are fairly simple and can share resources such as a worker pool.  If `redo` doesn't use `misc`, then in the rare occasions that `redo` gets a push or PR, it will need to wait for a worker to start in its worker pool, which can take a minute or two and delay getting feedback on the PR.

PR #32, opened 2019-10-24
Bug 1591243 - configure win2012r2, generic-worker, AWS images

djmitche:
WIP so far until I get an image that works -- this one starts but doesn't call registerWorker.

I'm not sure the workerConfig is correct - I'm not sure if workerConfig requires a `genericWorker` subkey.

https://bugzilla.mozilla.org/show_bug.cgi?id=1591243

### DIFF
--- a/generate/workers.py
+++ b/generate/workers.py
@@ -8,8 +8,63 @@ from tcadmin.resources import WorkerPool
 
 TYPES = {}
 
-GCP_PROVIDER = 'community-tc-workers-google'
-DEFAUlT_GCP_DOCKER_WORKER_IMAGE = "projects/taskcluster-imaging/global/images/docker-worker-gcp-googlecompute-2019-10-08t02-31-36z"
+AWS_PROVIDER = 'community-tc-workers-aws'
+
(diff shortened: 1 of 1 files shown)


djmitche:
It was incorrect, and the worker config requires a pretty deeply nested structure, and also requires lots of fields to be set that seem like they should be baked into the image (like paths on the image) -- but we can deal with that later.

generate/workers.py
```diff
@@ -105,3 +160,97 @@ def standard_gcp_docker_worker(*, image=None, diskSizeGb=50, privileged=False, *
...
+    """
+    Build a standard Win2012R2 worker instance in AWS
+    """
+    rv = base_aws_generic_worker_config(
+        imageIds=DEFAULT_AWS_WIN2012_GENERIC_WORKER_IMAGES,
+        instanceTypes={"m3.2xlarge": 1},
+        **cfg)
+
+    # instance type m3.2xlarge isn't available in this us-east-1a, so we filter
+    # out that zone
```
imbstack: ugh

PR #36, opened 2019-10-28
Bug 1592025 - assign public IPs to all hosts

djmitche:
Without this, the hosts can't communicate with the outside world, which fills them with sadness and prevents them from doing their jobs.


djmitche:
(This has already been applied..)

PR #38, opened 2019-10-28
Bug 1574666 - scopes for json-e deployments

djmitche:


config/projects.yml
```diff
@@ -75,6 +75,13 @@ taskcluster:
         - queue:route:notify.irc-channel.#taskcluster-bots.*
       to: repo:github.com/taskcluster/taskcluster:branch:master
 
+    - grant:
+        # pushes to json-e master can read secrets for deploying the site
+        - secrets:get:project/taskcluster/json-e-deploy
+        # ..and notify on failure
+        - queue:route:notify.email.taskcluster-notifications@mozilla.com.*
+      to: repo:github.com/taskcluster/json-e:branch:master
```
owlishDeveloper: In Tools, I saw that in current production this role also has `auth:aws-s3:read-write:taskcluster-raw-docs/json-e/` - do we need that one as well?

PR #39, opened 2019-10-28
Bug 1574661 - add firefoxreality project

djmitche:
Note that this does not add the win2012r2 worker for FirefoxRealityPC -- we can do that in a follow-up.

### DIFF
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -148,6 +148,52 @@ servo:
       minCapacity: 0
       maxCapacity: 2
 
+firefoxreality:
+  adminRoles:
+    - github-org-admin:MozillaReality
+  repos:
+    - github.com/MozillaReality/FirefoxReality
+    - github.com/MozillaReality/FirefoxRealityPC
+  workerPools:
+    ci-linux:
+      owner: nobody@mozilla.com
+      emailOnError: false
+      type: standard_gcp_docker_worker
+      minCapacity: 0
+      maxCapacity: 6
+  grants:
+    # General repo permissions
+    - grant:
+        - queue:create-task:highest:proj-firefoxreality/ci-linux
+        - queue:route:notify.email.fxr-releng@mozilla.com.*
+      to:
+        - repo:github.com/MozillaReality/FirefoxReality:*
+        - repo:github.com/MozillaReality/FirefoxRealityPC:*
+
+    # Firefox Reality
+    - grant:
+        - secrets:get:project/firefoxreality/fr/github-deploy-key
+        - secrets:get:project/firefoxreality/fr/key-password
+        - secrets:get:project/firefoxreality/fr/keystore-password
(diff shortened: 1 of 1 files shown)


config/projects.yml
```diff
@@ -148,6 +148,52 @@ servo:
...
 
+firefoxreality:
+  adminRoles:
+    - github-org-admin:MozillaReality
+  repos:
+    - github.com/MozillaReality/FirefoxReality
+    - github.com/MozillaReality/FirefoxRealityPC
+  workerPools:
+    ci:
+      owner: nobody@mozilla.com
```
petemoore: maybe pl**b@mozilla.com ?

config/projects.yml
```diff
@@ -148,6 +148,52 @@ servo:
...
       maxCapacity: 2
 
+firefoxreality:
+  adminRoles:
+    - github-org-admin:MozillaReality
+  repos:
+    - github.com/MozillaReality/FirefoxReality
+    - github.com/MozillaReality/FirefoxRealityPC
+  workerPools:
+    ci:
```
petemoore: I believe several worker pools will eventually be needed for CI (e.g. to support Windows 10, linux and possibly mac?). Maybe this should be `ci-linux` instead so that later e.g. `ci-win10` can be added.

config/projects.yml
djmitche: @phillip-lamb is that OK?

config/projects.yml
djmitche: good point, will revise..

config/projects.yml
djmitche: (we can land without this change and update later)

owlishDeveloper: oi wei, we're not supposed to commit these emails?

owlishDeveloper: I saw bhearsum's email commited

config/projects.yml
```diff
@@ -148,6 +148,52 @@ servo:
...
+  repos:
+    - github.com/MozillaReality/FirefoxReality
+    - github.com/MozillaReality/FirefoxRealityPC
+  workerPools:
+    ci-linux:
+      owner: nobody@mozilla.com
+      emailOnError: false
+      type: standard_gcp_docker_worker
+      minCapacity: 0
+      maxCapacity: 6
```
owlishDeveloper: how do we know this number?

config/projects.yml
djmitche: I think it's OK I'm just not 100% sure phillip is the right person :)

config/projects.yml
djmitche: I totally made that up... we can adjust upward later if there's complaints about pending tasks.

config/projects.yml
owlishDeveloper: oic :)

PR #40, opened 2019-10-29
Create docker-worker client for CI tests

walac:


config/projects.yml
```diff
@@ -120,6 +120,12 @@ taskcluster:
         - secrets:get:project/taskcluster/testing/docker-worker/pulse-creds
       to: repo:github.com/taskcluster/docker-worker:*
 
+  clients:
+    docker-worker/ci:
+      scopes:
+        - assume:repo:github.com/taskcluster/docker-worker:*
+      description: Client ID for docker-worker CI
```
djmitche: `description` is ignored, so this might be better written as a comment here

PR #41, opened 2019-10-29
Add taskcluster_yml_validator project

owlishDeveloper:


config/projects.yml
```diff
@@ -169,6 +169,21 @@ redo:
     - grant: queue:create-task:highest:proj-redo/ci
       to: repo:github.com/mozilla-releng/redo:*
 
+taskcluster_yml_validator:
+  adminRoles: []
```
djmitche: @milescrabill suggested adding marco's github identity to this (with a role like `login-identity:github/1038527|glandium` but with marco-c's user_id and username in it).

PR #45, opened 2019-10-31
Move taskcluster_yml_validator to the misc project

djmitche:
The underscores in the project name are incompatible with the GCP
provider.  This project only runs simple lint/CI, so misc is sufficient,
rather than renaming the project to `tcymlvalidator` or the like.


imbstack:
I could've sworn the gcp provider replaced underscores with dashes or something like this. This is a bug in gcp provider if this isn't accepted!

djmitche:
@imbstack would you rather stay with the distinct project, then, and try to fix the bug?

PR #49, opened 2019-10-31
Bug 1574668 - add a 'wpt' project

djmitche:
This project just has its own (dedicated) docker-worker worker pool.  As configured, this will run in GCP, but the effect should be pretty similar to the existing AWS worker type.

We can add to `adminRoles` as necessary -- if there's individual GitHub users who should have access, or GitHub teams, or all org admins of the wpt org.


djmitche:
& applied..

PR #52, opened 2019-11-01
replace redo project with releng project

milescrabill:
- the redo project is a releng project, but there are others, i.e. maven-lambda
- this lets us share the worker-pool across those projects

- [x] delete the scopes/roles for `redo` after replacing the worker-pool in its repo


milescrabill:
I've applied an intermediate configuration that will let both redo and maven-lambda use the `proj-releng` worker-pool

PR #55, opened 2019-11-02
Add managed SGs for SSH and RDP access

djmitche:


djmitche:
This also adds the AWS account ID, which is necessary for sharing AMIs with this account.

PR #56, opened 2019-11-02
Allow admins use of all providers

djmitche:
Rather than `*`, this allows just the existing providers.  That way, if we later add a project-specific provider, all admins won't have access to it.


djmitche:
Once https://github.com/taskcluster/tc-admin/pull/50 lands, the null-provider bit won't be necessary, but doesn't hurt.

PR #60, opened 2019-11-03
Grant scopes for IRC and Treeherder routes to Servo admins

SimonSapin:
Theeherder scopes were previously granted through https://tools.taskcluster.net/auth/roles/project%3Aservo%3Agrants%2Ftreeherder. I couldn’t find what role grants IRC scopes on taskcluster.net, though.


djmitche:
(applied)

PR #62, opened 2019-11-03
Grant `queue:scheduler-id:taskcluster-github` to servo

djmitche:
...to support making hooks that use this schedulerId


djmitche:
(I'm not sure what to do about schedulerId.. it just seems to be awkward in a bunch of ways.  Which is why I didn't give this to every project, just to servo)

PR #65, opened 2019-11-04
Grant project admins worker-related scopes (workerGroupId proj-<..>)

djmitche:
(already applied)


SimonSapin:
Is `assume:worker-pool:<something>` also needed?

djmitche:
> Is `assume:worker-pool:<something>` also needed?

Good point.  Added!

PR #68, opened 2019-11-04
Add sentry scopes for DS macos-light workers

djmitche:


djmitche:
(already applied)

djmitche:
withdrawn :)

PR #69, opened 2019-11-04
update worker-pool role, and use it and worker-id roles in DS clients

djmitche:


djmitche:
It does, [here](https://github.com/mozilla/community-tc-config/pull/69/files#diff-4330d89b51fcdb55e8375a442c4882eaR120)

PR #81, opened 2019-11-05
Grant task management to Servo through project admin role instead of github team role

SimonSapin:
It’s not terribly important. Project admin is what I’ve been granting to laptop / CLI clients.


djmitche:
(applied)

PR #86, opened 2019-11-06
Use n2-highcpu-32 for DeepSpeech

lissyx:


lissyx:
Also, I don't yet have setup `tc-admin`, can this be deployed for me?

SimonSapin:
@lissyx Note that `n2-*` machine are only available in some GCP regions / zones, and even then availability is limited. For Servo I’ve switched to `n1-*`, though I haven’t looked much into the details.

lissyx:
Good to know, but the variability **and** the increase in build time is really painful :/

djmitche:
I think they're in only one region, and only 24 CPUs allowed per project.  Let's look at this next week -- we may be able to request a limit increase from Google.

lissyx:
Duh, that's a lot of limitations all of a sudden...

djmitche:
Google only very recently announced the n2's.  I think they're still racking them up..

djmitche:
@ccooper has more data, i think..

config/projects.yml
```diff
@@ -443,7 +443,7 @@ deepspeech:
       type: standard_gcp_docker_worker
       minCapacity: 0
       maxCapacity: 96
-      machineType: "zones/{zone}/machineTypes/n1-standard-32"
+      machineType: "zones/{zone}/machineTypes/n2-highcpu-32"
```
ccooper: n2s are available now, but only in certain regions and zones: https://cloud.google.com/compute/docs/regions-zones/#available

You could potentially use them, but you'd need to be explicit about the zone(s) in your patch.

lissyx: Thanks! That page is hard to decipher :[

djmitche:
Looks like this is stalled out?

lissyx:
it did not, I was waiting to let the dust settle and have some more feedback on this zone thing.

lissyx:
So far we don't need this anymore for the moment.

PR #90, opened 2019-11-06
Admins get access to private/docker-worker artifacts

djmitche:
(already applied)


djmitche:
Yes, and the particular artifact we put there is `shell.html`, which allows access to the interactive shell.

PR #91, opened 2019-11-06
Add privileged to relman worker

imbstack:
This should fix bug 1593749

- [x] this is applied


imbstack:
They already had the scopes actually. I just forgot this part

PR #96, opened 2019-11-06
Make relman externally managed

imbstack:
This is to support bugbug deploying its own hooks


imbstack:
well... shoot.

imbstack:
I thought it just wouldn't delete things it didn't know about

djmitche:
That might just be a bug where we should add a manage regexp for something.

PR #98, opened 2019-11-07
Reconfigure relman to have correct roles/secrets and be simpler

imbstack:


djmitche:
..except some YAML lint ot clean up

PR #102, opened 2019-11-07
Bug 1587554 - add generic-worker roles/clients/workerpools

petemoore:
See [bug 1587554](https://bugzil.la/1587554) for context. Thanks!

### DIFF
--- a/config/imagesets.yml
+++ b/config/imagesets.yml
@@ -1,0 +1,59 @@
+docker-worker:
+  worker-implementation: docker-worker
+  gcp:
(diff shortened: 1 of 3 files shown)


petemoore:
Oh just realised I didn't add the AWS/GCP image references.... Let me do that now!

petemoore:
OK, looking at the code I see I need to do a bit of work here to create an appropriate function with config properties that can be included in `projects.yml`, so this is a little bit more involved so I'll leave it until the morning when I have a fresh head, as it is a little late for me to start on that now. I'll aim to just have a single generic worker function, and suitable properties to distinguish platform/privilege levels etc.

petemoore:
I'm replacing this PR with smaller ones.

PR #105, opened 2019-11-08
Bump maxruntime for jcristau-fx-release-metrics

jcristau:
The task seems to need slightly longer than on taskcluster.net


djmitche:
(applied)

PR #106, opened 2019-11-08
Add one-off-add-ons

owlishDeveloper:


owlishDeveloper:
duplicating from slack: that's already granted to `repo:github.com/mozilla/*` so I'm just merging

PR #107, opened 2019-11-11
Bug 1595464 - Add libmozevent to projects

marco-c:


marco-c:
"tc-admin diff" is failing on master too, I filed #108.

djmitche:
Applied, with an empty secret.

PR #111, opened 2019-11-11
Bug 1595416 - add rerun, etc. for relman admins

djmitche:


imbstack:
linter seems to be yelling but it lgtm

PR #114, opened 2019-11-12
Bug 1595589 - Add all scopes required by bugbug hooks

marco-c:


config/projects.yml
```diff
@@ -549,12 +549,20 @@ relman:
...
+      to: repo:github.com/mozilla/bugbug:*
+    # The build scopes for the tag will come from the previous rule.
     - grant: assume:project:relman:bugbug/deploy
       to: repo:github.com/mozilla/bugbug:tag:*
     - grant:
-        - "queue:route:notify.email.release-mgmt-analysis@mozilla.com.on-failed"
-        - "queue:route:notify.irc-channel.#bugbug.on-failed"
-        - "queue:route:index.project.relman.bugbug.data-pipeline-start"
+        - assume:project:relman:bugbug/build
+        - assume:project:relman:bugbug/deploy
```
marco-c: Is the "circular dependency" acceptable?

config/projects.yml
```diff
@@ -549,12 +549,20 @@ relman:
...
+        - hooks:trigger-hook:project-relman/bugbug*
+        - queue:route:notify.email.release-mgmt-analysis@mozilla.com.on-failed
+        - queue:route:notify.irc-channel.#bugbug.on-failed
+        - queue:route:index.project.relman.bugbug.*
+        - queue:route:index.project.relman.bugbug_annotate.*
```
marco-c: Would a merge of the two, "queue:route:index.project.relman.bugbug*", work?

config/projects.yml
```diff
@@ -549,12 +549,20 @@ relman:
         - queue:route:project.relman.bugbug.deploy_ending.*
```
marco-c: Do we need "queue:route:project.relman.bugbug.deploy_ending" and "queue:route:project.relman.bugbug.deploy_ending.*"? In bugbug I have https://github.com/mozilla/bugbug/blob/da439431d650e028c9f40d6e2d82ff3dcec7f4a9/.taskcluster.yml#L321 and https://github.com/mozilla/bugbug/blob/da439431d650e028c9f40d6e2d82ff3dcec7f4a9/infra/taskcluster-hook-pipeline-start.json#L5.

config/projects.yml
djmitche: it would!

config/projects.yml
djmitche: `deploy_ending*` should do, then.

config/projects.yml
djmitche: If the auth service doesn't reject it, it's OK.

config/projects.yml
marco-c: OK, done.

config/projects.yml
marco-c: OK, done.

config/projects.yml
djmitche: `taskcluster.exceptions.TaskclusterRestFailure: Roles may not contain dependency cycles, found: 'assume:hook-id:project-relman/bugbug*' -> 'assume:project:relman:bugbug/deploy' -> 'assume:hook-id:project-relman/bugbug*'`

djmitche:
Sorry, I accidentally reverted this since I only clicked "merge" once in the UI.  I shouldn't do things while in meetings!

djmitche:
(it's applied now)

PR #115, opened 2019-11-12
add temporary access for tomprince to set secrets for relman

djmitche:


djmitche:
(applied -- let's leave this open to unapply later)

PR #116, opened 2019-11-12
Define a small and large machine type

marco-c:


djmitche:
(sorry, it's late for you and that result is pretty obvious so I'll assume "yes")

PR #118, opened 2019-11-12
allow team 'smoketesters' to run smoketests

djmitche:
See taskcluster/taskcluster#1859.


djmitche:
I think we can get away without that.. if so, we can add it.

PR #126, opened 2019-11-19
Bug 1583990 - manage secrets with values from passwordstore

djmitche:
I sort of threw this together.  If this is the wrong approach, let me know.

For the most part (except for our azure secret), this leaves existing secrets externally managed.  That's probably best for most projects other than taskcluster, anyway.

What this will really help with is worker-pool secrets -- those are now managed automatically.

### DIFF
--- a/generate/secret_values.py
+++ b/generate/secret_values.py
@@ -1,0 +1,48 @@
+# -*- coding: utf-8 -*-
+
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at http://mozilla.org/MPL/2.0/.
+
+import subprocess
+import sys
+import yaml
+
+PASSWORDSTORE_NAME = "community-tc/secret-values.yml"
+
+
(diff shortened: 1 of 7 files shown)


generate/secret_values.py
```diff
@@ -0,0 +1,48 @@
...
+PASSWORDSTORE_NAME = "community-tc/secret-values.yml"
+
+
+class SecretValues:
+    """A container for secret values.  This is loaded only when --with-secrets,
+    and requires `passwordstore` support."""
+
+    def __init__(self):
+        print("Fetching secrets with passwordstore", file=sys.stderr)
+        secret_values_yml = subprocess.check_output(["pass", PASSWORDSTORE_NAME])
```
imbstack: What will this do if I need to enter a passphrase?

djmitche: That's a good question -- I didn't test it since I yubikey, but I think it should work -- it doesn't intercept stdin, and the prompt should go to stderr.  Can you test it out?

imbstack: It seems to work fine!

PR #129, opened 2019-11-20
Allow GitHub task to forward the 'statuses' and 'checks' routes

SimonSapin:
… in order to work around current tc-github limitations.

### DIFF
--- a/config/grants.yml
+++ b/config/grants.yml
@@ -163,3 +163,15 @@
 
     # for relman
     - project-admin:relman
+
+# One of these routes is added to each task created in reaction to a GitHub event.
+# If such a task in turn creates other tasks, it need to add the same route in order to work around
+# https://bugzilla.mozilla.org/show_bug.cgi?id=1548781 or
+# https://bugzilla.mozilla.org/show_bug.cgi?id=1533235
+- grant:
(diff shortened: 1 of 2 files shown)


config/grants.yml
```diff
@@ -163,3 +163,15 @@
...
+# If such a task in turn creates other tasks, it need to add the same route in order to work around
+# https://bugzilla.mozilla.org/show_bug.cgi?id=1548781 or
+# https://bugzilla.mozilla.org/show_bug.cgi?id=1533235
+- grant:
+    # Route names match configuration at
+    # https://github.com/taskcluster/taskcluster/blob/21f257dc8/services/github/config.yml#L13-L14
+    - queue:route:checks
+    - queue:route:statuses
+  to:
+    - repo:github.com/*
```
djmitche: Since this is a workaround, I'd like to limit this to a more narrow set of repositories (that we revisit when the workaround is no longer required).

SimonSapin: Well, I’m not aware of any plan for how we could make this not required, or anyone working on this. But if you prefer that I don’t mind, I’ve limited this to `servo/servo` (and other repos can be added as needed)

PR #130, opened 2019-11-22
Add scopes to trigger pulse notifications to bugbug hooks

marco-c:


config/projects.yml
```diff
@@ -553,8 +553,6 @@ relman:
         - hooks:modify-hook:project-relman/bugbug-checks
         - hooks:modify-hook:project-relman/bugbug-classify-patch
         - hooks:modify-hook:project-relman/bugbug-test-select
-        - queue:route:notify.pulse.route.project.relman.bugbug.*
-        - queue:route:project.relman.bugbug.*
```
marco-c: I'm removing them from here, since it has assume:hook, so they are inherited

djmitche:
applied

PR #132, opened 2019-11-22
Raise maxCapacity for Servo’s worker pools

SimonSapin:
We’ve added a number of Linux tasks to CI recently, and have occasionally reached the previous maximum since.


djmitche:
applied

PR #133, opened 2019-11-26
Bug 1587554 - part 1: add generic-worker CI roles and clients

petemoore:
I've decided to split this up into several bite size parts that can be landed independently.

This first part adds generic-worker CI roles and clients.

The next part is a little more disruptive, but this part shouldn't have any surprises.

Thanks!

### DIFF
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -186,14 +186,14 @@ taskcluster:
         - generic-worker:os-group:test-provisioner/*
         - generic-worker:run-as-administrator:test-provisioner/*
         - queue:cancel-task:test-scheduler/*
-        - queue:claim-work:test-provisioner/*
+        - assume:worker-pool:test-provisioner/*
         - queue:create-artifact:public/*
         - queue:create-task:test-provisioner/*
         - queue:get-artifact:SampleArtifacts/_/X.txt
         - queue:get-artifact:SampleArtifacts/_/non-existent-artifact.txt
         - queue:get-artifact:SampleArtifacts/b/c/d.jpg
         - queue:resolve-task
-        - queue:worker-id:test-worker-group/*"
+        - assume:worker-id:test-worker-group/*"
       to: project:taskcluster:generic-worker-tester
 
     - grant:
@@ -217,8 +217,8 @@ taskcluster:
     # Client for workerpool proj-taskcluster/gw-ci-macos-staging.
     generic-worker/ci-macos-staging:
       scopes:
-        - queue:claim-work:proj-taskcluster/gw-ci-macos-staging
-        - queue:worker-id:proj-taskcluster/*
+        - assume:worker-pool:proj-taskcluster/gw-ci-macos-staging
+        - assume:worker-id:proj-taskcluster/*
     # A client whose clientId and accessToken are stored in taskcluster secret
     # `repo:github.com/taskcluster/generic-worker` and used by commands in
     # generic-worker's .taskcluster.yml to create production tasks that are
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -179,7 +179,6 @@ taskcluster:
         - auth:create-client:project/taskcluster:generic-worker-tester/TestReclaimCancelledTask
         - auth:create-client:project/taskcluster:generic-worker-tester/TestResolveResolvedTask
         - auth:sentry:generic-worker-tests
-        - auth:webhooktunnel
         - generic-worker:cache:banana-cache
         - generic-worker:cache:devtools-app
         - generic-worker:cache:test-modifications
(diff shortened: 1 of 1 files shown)


config/projects.yml
```diff
@@ -173,13 +173,69 @@ taskcluster:
...
+        - auth:sentry:generic-worker-tests
+        - auth:webhooktunnel
+        - generic-worker:cache:banana-cache
+        - generic-worker:cache:devtools-app
+        - generic-worker:cache:test-modifications
+        - generic-worker:cache:unknown-issuer-app-cache
+        - generic-worker:os-group:test-provisioner/*
+        - generic-worker:run-as-administrator:test-provisioner/*
+        - queue:cancel-task:test-scheduler/*
+        - queue:claim-work:test-provisioner/*
```
djmitche: Ideally this `test-provisioner` stuff would become `proj-taskcluster/gw-*` or something of the sort..

config/projects.yml
```diff
@@ -173,13 +173,69 @@ taskcluster:
     - grant: assume:project:taskcluster:smoketests
       to: github-team:taskcluster/smoketesters
 
+    - grant:
+        - assume:worker-id:test-worker-group/test-worker-id
+        - assume:worker-type:test-provisioner/*
+        - auth:create-client:project/taskcluster:generic-worker-tester/TestReclaimCancelledTask
+        - auth:create-client:project/taskcluster:generic-worker-tester/TestResolveResolvedTask
+        - auth:sentry:generic-worker-tests
+        - auth:webhooktunnel
```
djmitche: This scope doesn't do anything anymore

config/projects.yml
```diff
@@ -173,13 +173,69 @@ taskcluster:
     - grant: assume:project:taskcluster:smoketests
       to: github-team:taskcluster/smoketesters
 
+    - grant:
+        - assume:worker-id:test-worker-group/test-worker-id
```
djmitche: It'd be nice to have this be `proj-taskcluster/gw-test-worker` or the like here.

config/projects.yml
```diff
@@ -173,13 +173,69 @@ taskcluster:
...
       scopes:
         - assume:project:taskcluster:smoketests
     docker-worker/ci:
       scopes:
         - assume:repo:github.com/taskcluster/docker-worker:*
+    # Client for workerpool proj-taskcluster/gw-ci-macos.
+    generic-worker/ci-macos:
+      scopes:
+        - queue:claim-work:proj-taskcluster/gw-ci-macos
+        - queue:worker-id:proj-taskcluster/*
```
djmitche: Best to use `assume:worker-pool:proj-taskcluster/gw-ci-macos` and `assume:worker-id:proj-taskcluster/*` here -- that gets the queue scopes plus secret scopes, etc.

config/projects.yml
petemoore: The `test-provisioner/*` worker pool is historically the worker pool that worker integration tests use, to ioslate them from production worker pools (much like we had the `garbage:*` secrets for non-production test secrets). The `proj-taskcluster/*` worker pools are the actual (production) worker pools we use for running taskcluster CI workers, rather than "fake" worker pools only used in worker tests, so it might be better for us not to grant permissions to integration tests to run tasks as administrator on those particular production worker pools.

config/projects.yml
petemoore: Thanks, I didn't realise that. I'll remove it. 👍

config/projects.yml
petemoore: See above - I think we are better continuing to use the dummy/test worker namespaces for integration tests so that they don't conflict/collide with production namespaces during integration test runs. I think at some point Jonas put together a namespace doc on the names we had agreed on for test workers, workerGroups, workerIds, provisionerIds, workerTypes etc - and I believe these names still conform to that doc (unfortunately I can't find the link now, or maybe it got lost in a doc restructuring project).

config/projects.yml
petemoore: 👍

config/projects.yml
djmitche: That makes sense..

config/projects.yml
```diff
@@ -173,13 +173,68 @@ taskcluster:
...
+    # Client for workerpool proj-taskcluster/gw-ci-macos.
+    generic-worker/ci-macos:
+      scopes:
+        - assume:worker-pool:proj-taskcluster/gw-ci-macos
+        - assume:worker-id:proj-taskcluster/*
+    # Client for workerpool proj-taskcluster/gw-ci-macos-staging.
+    generic-worker/ci-macos-staging:
+      scopes:
+        - queue:claim-work:proj-taskcluster/gw-ci-macos-staging
+        - queue:worker-id:proj-taskcluster/*
```
djmitche: assume: here, too :)

petemoore: Ah yes, thanks!

PR #134, opened 2019-11-26
Move MozillaSecurity repos to a separate project: fuzzing

jschwartzentruber:
I think this makes sense since I'm adding secrets and fuzzing team may be doing more than CI in the near future.


djmitche:
applied just now

PR #135, opened 2019-11-27
Bug 1587554 part 2

petemoore:
This second part restructures the code and config a little to provide additional metadata for machine images.


petemoore:
I meant to create this as a draft pull request, and can't see a way to change it, so I'll create a new separate draft pull request.

PR #136, opened 2019-11-27
Bug1587554 part 2: restructure code/config to use image sets

petemoore:
The aim of this second PR is to restructure the organisation of the code and configs a little to enable additional metadata to be associated to machine images, and provide a slightly more generic means for worker config to be associated to worker pools.

### DIFF
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -31,6 +31,8 @@
 #     <worker-pool-name>:
 #       owner: ..
 #       emailOnError: ..
+#       instanceTypes: a dict, where each key is an instance type, and the
+#           value is the capacity per instance for that instance type.
 #       imageset: top level key from imagesets.yml
 #       cloud: cloud to deploy in ('aws' or 'gcp')
 #       ..: ..  # arguments to that function
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -93,10 +93,10 @@ taskcluster:
           allowPrivileged: true
 
     # gw-ci-* worker pools are for generic-worker CI
-    gw-ci-ubuntu-17-10:
+    gw-ci-ubuntu-18-04:
       owner: taskcluster-notifications+workers@mozilla.com
       emailOnError: true
-      imageset: generic-worker-ubuntu-17-10
+      imageset: generic-worker-ubuntu-18-04
       cloud: aws
       maxCapacity: 10
       workerConfig:
@@ -104,10 +104,10 @@ taskcluster:
           config:
             runTasksAsCurrentUser: true
 
-    gw-ci-ubuntu-17-10-staging:
+    gw-ci-ubuntu-18-04-staging:
       owner: taskcluster-notifications+workers@mozilla.com
       emailOnError: true
-      imageset: generic-worker-ubuntu-17-10-staging
+      imageset: generic-worker-ubuntu-18-04-staging
       cloud: aws
       maxCapacity: 10
       workerConfig:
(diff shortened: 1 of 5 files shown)


config/projects.yml
```diff
@@ -31,7 +31,8 @@
 #     <worker-pool-name>:
 #       owner: ..
 #       emailOnError: ..
-#       type: <name of a function in `generate/workers.py`)
+#       imageset: top level key from imagesets.yml
+#       cloud: cloud to deploy in ('aws' or 'gcp')
```
djmitche: need doc for `instanceTypes` too (it's not obvious what the values mean)

generate/workers.py
```diff
@@ -6,98 +6,62 @@
...
+     - secret - content of the `worker-pool:<workerPoolId>` secret (optional)
+    """
+    WORKER_IMPLEMENTATION_FUNCS[fn.__name__] = fn
     return fn
 
 
-def build_worker_pool(workerPoolId, cfg, secret_values):
+def merge(source, destination):
+    """
+    Recursively merges a source and destination dict
```
djmitche: maybe note that this overwrites arrays in source with arrays in dest?

generate/workers.py
```diff
@@ -6,98 +6,62 @@
...
+    CLOUD_FUNCS[fn.__name__] = fn
+    return fn
 
-    TYPES[fn.__name__] = fn
+
+def worker_implementation(fn):
+    """
+    Register a worker implementation generator. This function takes keyword arguments
+    based on the configuration in `projects.yml`, plus `secret_values`, an instance of
+    SecretValues
```
djmitche: It looks like this also takes `wp`?

generate/workers.py
```diff
@@ -6,98 +6,62 @@
...
+
+CLOUD_FUNCS = {}
+WORKER_IMPLEMENTATION_FUNCS = {}
+
+
+def cloud(fn):
     """
-    Register a worker pool generator.  This function takes keyword arguments based on the
+    Register a cloud config generator. This function takes keyword arguments based on the
     configuration in `projects.yml`, plus `secret_values`, an instance of SecretValues or
```
djmitche: It looks like this takes `image_set` too..

generate/workers.py
```diff
@@ -123,21 +87,42 @@ def build_worker_pool(workerPoolId, cfg, secret_values):
...
     """
-    Build a base config for a Google instance
+    Build a worker pool in Google.
 
-      minCapacity: minimum capacity to run at any time (default 0)
-      maxCapacity: maximum capacity to run at any time (required)
+      image: image name (defaults to the standard image)
+      diskSizeGb: boot disk size, in Gb (defaults to 50)
+      privileged: true if this worker should allow privileged tasks (default false)
+      ..in addition to kwargs from base_google_config
```
djmitche: I think this last parameter should not be documented here..

generate/workers.py
```diff
@@ -263,70 +281,56 @@ def base_aws_config(
...
-
-    # by default, deploy where there are images
-    if "regions" not in cfg and "imageIds" in cfg:
-        cfg["regions"] = list(cfg["imageIds"])
+@worker_implementation
+def generic_worker(image_set, wp, **cfg):
 
-    rv = base_aws_config(**cfg)
+    # hashed = hashlib.sha256(json.dumps(configs, sort_keys=True).encode("utf8")).hexdigest()
+    # generic_worker_config["deploymentId"] = hashed[:16]
```
djmitche: Let's remove or un-comment this..

PR #138, opened 2019-11-29
Bug 1587554 - m3.2xlarge isn't available in us-east-1f

ccooper:


owlishDeveloper:
Can we actually just ask Amazon what they have available where? 🤔

PR #140, opened 2019-12-03
Add dump_syms to projects

calixteman:


djmitche:
Looks like this is already deployed, too.

PR #144, opened 2019-12-03
Support worker-implementation specific scopes, shrink generic-worker configs, and configure deploymentId

petemoore:
This adds support for enabling `generic-worker` or `docker-worker` specific scopes, and cloud-specific scopes. This use case for this was wanting to give `auth:sentry:generic-worker` to all generic-worker workers.

### DIFF
--- a/generate/workers.py
+++ b/generate/workers.py
@@ -45,14 +45,20 @@ def worker_implementation(fn):
     return fn
 
 
-def merge(source, destination):
+def merge(*dicts):
     """
-    Returns deep merge of source and destination dicts. Array values inside
-    dicts are not merged. Values in source take precedence over values in
-    destination. Source and destination dicts are not altered.
+    Returns a new dict containing deep merge of dicts. Source dicts are not
+    altered. Array values inside dicts are not merged. Values in earlier dicts
+    take precedence over values in later dicts. At least two dicts required.
     """
-    result = copy.deepcopy(destination)
-    for key, value in source.items():
+
+    assert len(dicts) >= 2
+
+    if len(dicts) > 2:
+        return merge(dicts[0], merge(*dicts[1:]))
+
+    result = copy.deepcopy(dicts[1])
+    for key, value in dicts[0].items():
         if isinstance(value, dict):
             # get node or create one
             node = result.setdefault(key, {})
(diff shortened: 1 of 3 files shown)


petemoore:
It occurred to me after writing this, we could maybe do this with roles, e.g.

* `worker-implementation:generic-worker`
* `worker-implementation:generic-worker`
* `cloud-provider:aws`
* `cloud-provider:gcp`

and then e.g. add the scope `assume:worker-implementation:generic-worker` to generic-worker workers etc...

@djmitche Let me know if you prefer that, and I can add the `worker-implementation` and `cloud-provider` roles to `grants.yml` and then update `workers.py` to return the assume scopes instead.

generate/workers.py
```diff
@@ -284,48 +309,63 @@ def aws(
...
+    # the first launchConfig, regardless of the region they are in.
+    hashedConfig = hashlib.sha256(
+        json.dumps(wp["config"], sort_keys=True).encode("utf8")
+    ).hexdigest()
+
+    for launchConfig in wp["config"]["launchConfigs"]:
+        launchConfig["workerConfig"]["genericWorker"]["config"][
+            "deploymentId"
+        ] = hashedConfig[:16]
+
```
imbstack: Could we explicitly exclude capacity from this?

generate/workers.py
```diff
@@ -32,41 +33,60 @@ def worker_implementation(fn):
...
      - config - passed to worker-manager
      - secret - content of the `worker-pool:<workerPoolId>` secret (optional)
+     - scopes - any additional scopes required for workers with this
+                implementation
     """
     WORKER_IMPLEMENTATION_FUNCS[fn.__name__] = fn
     return fn
 
 
 def merge(source, destination):
```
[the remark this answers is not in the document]
petemoore: Good idea! I've updated the function to be variadic. 👍

generate/workers.py
petemoore: I changed the hash to be based just on the launchConfigs section rather than the full config, so maxCapacity and minCapacity are no longer included. Funnily enough the docs already said it was a hash of the launch config, but the code was hashing the entire config, so the docs were ok. However, I reworded them to be a little bit clearer at the same time. Thanks!

PR #145, opened 2019-12-05
Add client for manually configured proj-taskcluster/gw-ci-windows10-amd64

petemoore:
This is a stop gap for generic-worker CI until we have an azure provider.

### DIFF
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -276,10 +276,10 @@ taskcluster:
       scopes:
         - assume:worker-pool:proj-taskcluster/gw-ci-raspbian-stretch
         - assume:worker-id:proj-taskcluster/*
-    # Client for workerpool proj-taskcluster/gw-ci-windows10-amd64. This is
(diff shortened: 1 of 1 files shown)


config/projects.yml
```diff
@@ -276,6 +276,14 @@ taskcluster:
       scopes:
         - assume:worker-pool:proj-taskcluster/gw-ci-raspbian-stretch
         - assume:worker-id:proj-taskcluster/*
+    # Client for workerpool proj-taskcluster/gw-ci-windows10-amd64. This is
+    # currently a manually configured worker pool with a single worker, running
+    # in Azure. When we have an azure provider, we can terminate this worker,
+    # delete this client, and use worker manager provisioned instances instead.
```
djmitche: Just to correct a difference of terminology here: it's not actually a worker pool (that would be a pool using the static provider, in which case it wouldn't require a client).  It's a standalone worker for that taskQueueId.

PR #146, opened 2019-12-06
[Bug 1601810] Allow relman to notify

imbstack:
These clients already exist so I just imported them into this config. Let me know if you'd rather manually configure them, @marco-c (I think maybe client support didn't exist here when we first set up relman).

This also gives all of the "admin types" broad notify permissions.

### DIFF
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -640,6 +640,7 @@ relman:
     bugbug/code-review-testing:
       scopes:
         - hooks:trigger-hook:project-relman/bugbug-*
+        - notify:email:*
   workerPools:
     ci:
       owner: mcastelluccio@mozilla.com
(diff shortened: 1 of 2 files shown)


config/projects.yml
```diff
@@ -632,6 +632,14 @@ relman:
...
     - github.com/mozilla/dump_syms:*
     - github.com/mozilla/libmozevent:*
+  clients:
+    bugbug/code-review-production:
+      scopes:
+        - hooks:trigger-hook:project-relman/bugbug-*
+        - notify:email:*
+    bugbug/code-review-testing:
+      scopes:
+        - hooks:trigger-hook:project-relman/bugbug-*
```
marco-c: Could you add notify:email:* to the testing client too?

marco-c:
> These clients already exist so I just imported them into this config. Let me know if you'd rather manually configure them, @marco-c (I think maybe client support didn't exist here when we first set up relman).

Yes, thanks, handling them via this repo is much better!

PR #148, opened 2019-12-06
Move imagesets from generic-worker repo into community-tc-config

petemoore:
This is a wholesale move over of the [/imagesets directory](https://github.com/taskcluster/generic-worker/tree/90832377da4925d2ac3a181da08fe977fe31861c/imagesets) in the generic-worker repo.

Once this PR lands, I'll delete it from the generic-worker repo.

It contains a single shell script which can be used to deploy images to aws or gcp, plus a directory for each image set, containing some config settings plus a bootstrap script for bootstrapping the image.


imagesets/README.md
```diff
@@ -0,0 +1,10 @@
+### Building an image set
+
+To build an image set:
+
+  * AWS only: Sign in with 2FA and export env vars `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_SESSION_TOKEN`
+  * GCP only: `export GCP_PROJECT=<your google cloud project of choice to deploy image into>`
```
djmitche: I think there's only one project that's used here, right?

imagesets/firefoxreality-1-win10-64/aws_owners
```diff
@@ -0,0 +1,2 @@
+self
+amazon
```
imbstack: Not blocking landing on this but can we file a followup bug to document what each of these files mean and how we should use them?

imagesets/README.md
petemoore: That's a good question - I'm not sure.

@imbstack is there a particular GCP project we should deploy to?

imagesets/firefoxreality-1-win10-64/aws_owners
petemoore: Sure, I'll follow up with another PR, that's a good idea.

I'll also see if I can get `tc-admin` to call the script directly unless you specify `--no-image-creation` (or similar) if it detects any changes (based on a hash digest of the content). This would save us needing to manually copy/paste ami ids around etc, and make the workflow considerably easier.

Maybe we can also auto-generate the link to the image set definition in the [`workerTypeMetaData`](https://github.com/taskcluster/generic-worker/blob/v16.5.5/usage.go#L282-L288) config section too. That is also manually managed at the moment.

PR #150, opened 2019-12-09
Add support for any codecov token and pypi credentials being fetched from fuzzing CI.

jschwartzentruber:


config/projects.yml
```diff
@@ -546,7 +546,9 @@ fuzzing:
   grants:
     - grant:
         - queue:create-task:highest:proj-fuzzing/ci
-        - secrets:get:project/fuzzing/github-deploy-key
+        - secrets:get:project/fuzzing/*-deploy-key
+        - secrets:get:project/fuzzing/*-codecov
```
petemoore:
```suggestion
        - secrets:get:project/fuzzing/codecov-*
```

config/projects.yml
```diff
@@ -546,7 +546,9 @@ fuzzing:
   grants:
     - grant:
         - queue:create-task:highest:proj-fuzzing/ci
-        - secrets:get:project/fuzzing/github-deploy-key
+        - secrets:get:project/fuzzing/*-deploy-key
```
petemoore:
```suggestion
        - secrets:get:project/fuzzing/deploy-key-*
```

config/projects.yml
jschwartzentruber: No problem.

config/projects.yml
jschwartzentruber: I realized this doesn't actually make sense. I can just use a common deploy key for ci in all the repos we want to support. Since the grant is on MozillaSecurity/*, if one leaks, they all leak anyways.

PR #153, opened 2019-12-09
Bug 1558532 - use worker-runner for generic-worker-ubuntu-18-04 image…

djmitche:
… sets

### DIFF
--- a/imagesets/generic-worker-ubuntu-18-04-staging/bootstrap.sh
+++ b/imagesets/generic-worker-ubuntu-18-04-staging/bootstrap.sh
@@ -94,20 +94,9 @@ User=root
 RequiredBy=graphical.target
 EOF
 
-cloud="%MY_CLOUD%"
-if [ -z "$providerType" ]; then
-    if [ "$cloud" = "aws" ]; then
-        providerType=aws
-    elif [ "$cloud" = "gcp" ]; then
-        providerType=google
-    fi
-fi
-if [ -z "$providerType" ]; then
-    echo "No provider type for cloud $cloud"
-fi
 cat > /etc/start-worker.yml <<EOF
 provider:
-    providerType: $providerType
+    providerType: %MY_CLOUD%
 worker:
     implementation: generic-worker
     path: /usr/local/bin/generic-worker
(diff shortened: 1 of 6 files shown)


djmitche:
This is now working.  Pete, can you have a look and let me know what you think?

I've only updated the AMIs for staging, but the scripts update both.  I can go back and install the same AMIs for both.

This can't land until https://bugzilla.mozilla.org/show_bug.cgi?id=1602960, at least, as it fails without that.

imagesets/generic-worker-ubuntu-18-04-staging/bootstrap.sh
```diff
@@ -74,7 +76,28 @@ mkdir -p /var/local/generic-worker
...
+if [ -z "$providerType" ]; then
+    if [ "$cloud" = "aws" ]; then
+        providerType=aws
+    elif [ "$cloud" = "gcp" ]; then
+        providerType=google
+    fi
+fi
+if [ -z "$providerType" ]; then
+    echo "No provider type for cloud $cloud"
+fi
```
petemoore: Let's just align `imageset.sh` to use keys `aws|google` instead of `aws|gcp` and then lines 81-91 aren't needed and we reduce complexity. Then update line 94 to `providerType: %MY_CLOUD%`.

djmitche: That does simplify things.

djmitche:
I re-generated images and [ran a task](https://community-tc.services.mozilla.com/tasks/UJk_yXW7SoKSFguxOjnkvQ) successfully.

petemoore:
It is a bit late now, but I realised that `aws|gcp` or `amazon|google` probably make more sense, but I don't suppose it really matters. :-)

djmitche:
> It is a bit late now, but I realised that aws|gcp or amazon|google probably make more sense, but I don't suppose it really matters. aws|google is a bit inconsistent, but never mind. :-)

Yeah, we sort of missed that in naming the providers in worker-manager, but we can at least make this consistent with it :(

djmitche:
Reminder to self:
> This can't land until https://bugzilla.mozilla.org/show_bug.cgi?id=1602960, at least, as it fails without that.

djmitche:
The major change since your last review is in the startup process -- this switches to a systemd-based startup, and has the cool advantage of including the startup logging in the AWS console:

Screenshot_2019-12-19 Instances EC2 Management Console

PR #156, opened 2019-12-10
Setup initial CI for Fuzzing

La0:
- CI secret
- ~docker privileged mode~
- simple hook

### DIFF
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -582,7 +582,7 @@ fuzzing:
               mkdir ~/.ssh
               ssh-keyscan github.com > ~/.ssh/known_hosts
               cd ~/.ssh
-              curl -L http://taskcluster/api/secrets/v1/secret/$DEPLOY_KEY | python -c 'import json, sys; a = json.load(sys.stdin); open("id_ed25519", "w").write(a["secret"]["key"])'
+              curl -L "${TASKCLUSTER_PROXY_URL}/api/secrets/v1/secret/${DEPLOY_KEY}" | python -c 'import json, sys; a = json.load(sys.stdin); open("id_ed25519", "w").write(a["secret"]["key"])'
               chmod 0400 ~/.ssh/id_ed25519
               git clone --quiet git@github.com:MozillaSecurity/fuzzing-tc-config.git /src
               pip install --disable-pip-version-check --no-cache-dir --quiet /src
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -581,8 +581,7 @@ fuzzing:
               apt-get install -qq openssh-client curl git > /dev/null
               mkdir ~/.ssh
               ssh-keyscan github.com > ~/.ssh/known_hosts
-              cd ~/.ssh
-              curl -L "${TASKCLUSTER_PROXY_URL}/api/secrets/v1/secret/${DEPLOY_KEY}" | python -c 'import json, sys; a = json.load(sys.stdin); open("id_ed25519", "w").write(a["secret"]["key"])'
+              curl -L "$TASKCLUSTER_PROXY_URL/api/secrets/v1/secret/$DEPLOY_KEY" | python -c 'import json, os, sys; a = json.load(sys.stdin); open(os.path.expanduser("~/.ssh/id_ed25519"), "w").write(a["secret"]["key"])'
               chmod 0400 ~/.ssh/id_ed25519
               git clone --quiet git@github.com:MozillaSecurity/fuzzing-tc-config.git /src
               pip install --disable-pip-version-check --no-cache-dir --quiet /src
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -549,14 +549,63 @@ fuzzing:
       cloud: gcp
       minCapacity: 0
       maxCapacity: 10
+    decision:
+      owner: fuzzing+taskcluster@mozilla.com
+      emailOnError: false
+      imageset: docker-worker
+      cloud: gcp
+      minCapacity: 0
+      maxCapacity: 10
+  hooks:
+    decision:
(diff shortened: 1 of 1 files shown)


config/projects.yml
```diff
@@ -546,14 +546,59 @@ fuzzing:
...
+            - -c
+            - >-
+              apt-get update && apt-get install -y openssh-client curl git &&
+              mkdir ~/.ssh &&
+              ssh-keyscan github.com > ~/.ssh/known_hosts &&
+              curl -L ${TASKCLUSTER_SECRET} | python -c 'import json, sys; a = json.load(sys.stdin); open(".ssh/id_ed25519", "w").write(a["secret"]["key"])' &&
+              chmod 0400 .ssh/id_ed25519 &&
+              git clone --quiet git@github.com:MozillaSecurity/fuzzing-tc-config.git /src &&
+              pip install /src &&
+              fuzzing-decision
```
petemoore: nit: to avoid all the `&&`'s you can rewrite as:

```
- /bin/bash
- '-ce'
- |
  apt-get update
  apt-get install -y openssh-client curl git
  mkdir ~/.ssh
  .....
```

I think this looks a bit nicer, but isn't critical.

La0: TIL !

config/projects.yml
```diff
@@ -549,14 +549,63 @@ fuzzing:
...
+      schedule:
+        - "0 0 */1 * * *"
+      task:
+        provisionerId: proj-fuzzing
+        workerType: decision
+        payload:
+          image: python:3.7-slim
+          features:
+            taskclusterProxy: true
+          environment:
```
petemoore:
```suggestion
          env:
```

(see [payload docs](https://github.com/taskcluster/docker-worker/blob/296b21cf8668b33fcf0869a2c4355323427d8c3a/schemas/v1/payload.json#L163-L167))

config/projects.yml
```diff
@@ -549,14 +549,63 @@ fuzzing:
...
+          command:
+            - /bin/bash
+            - -cxe
+            - |
+              apt-get update
+              apt-get install -qq openssh-client curl git > /dev/null
+              mkdir ~/.ssh
+              ssh-keyscan github.com > ~/.ssh/known_hosts
+              cd ~/.ssh
+              curl -L http://taskcluster/api/secrets/v1/secret/$DEPLOY_KEY | python -c 'import json, sys; a = json.load(sys.stdin); open("id_ed25519", "w").write(a["secret"]["key"])'
```
petemoore: This works, but the following might be better (in case taskcluster proxy url changes at some point).

```suggestion
              curl -L "${TASKCLUSTER_PROXY_URL}/api/secrets/v1/secret/${DEPLOY_KEY}" | python -c 'import json, sys; a = json.load(sys.stdin); open("id_ed25519", "w").write(a["secret"]["key"])'
```

(see [`TASKCLUSTER_PROXY_URL` docs](https://community-tc.services.mozilla.com/docs/manual/design/env-vars#taskcluster_proxy_url))

config/projects.yml
```diff
@@ -549,14 +549,63 @@ fuzzing:
...
+            DEPLOY_KEY: project/fuzzing/deploy-fuzzing-tc-config
+            TASKCLUSTER_SECRET: project/fuzzing/decision
+          command:
+            - /bin/bash
+            - -cxe
+            - |
+              apt-get update
+              apt-get install -qq openssh-client curl git > /dev/null
+              mkdir ~/.ssh
+              ssh-keyscan github.com > ~/.ssh/known_hosts
```
petemoore: Interesting! Do we get an error without this?

config/projects.yml
```diff
@@ -549,14 +549,63 @@ fuzzing:
...
+              cd ~/.ssh
+              curl -L http://taskcluster/api/secrets/v1/secret/$DEPLOY_KEY | python -c 'import json, sys; a = json.load(sys.stdin); open("id_ed25519", "w").write(a["secret"]["key"])'
+              chmod 0400 ~/.ssh/id_ed25519
+              git clone --quiet git@github.com:MozillaSecurity/fuzzing-tc-config.git /src
+              pip install --disable-pip-version-check --no-cache-dir --quiet /src
+              fuzzing-decision
```
petemoore: Will `fuzzing-decision` be in `PATH`?

config/projects.yml
```diff
@@ -549,14 +549,63 @@ fuzzing:
...
+            TASKCLUSTER_SECRET: project/fuzzing/decision
+          command:
+            - /bin/bash
+            - -cxe
+            - |
+              apt-get update
+              apt-get install -qq openssh-client curl git > /dev/null
+              mkdir ~/.ssh
+              ssh-keyscan github.com > ~/.ssh/known_hosts
+              cd ~/.ssh
```
petemoore: Perhaps better to not `cd ~/.ssh` here, but to explicitly write to `~/.ssh/id/id_ed25519` in line below? Also because this is the last `cd` before running `fuzzing`, so the working directory of the `fuzzing` process will be `~/.ssh`

config/projects.yml
La0: Yeap, it's a console script from the python package i'm building in https://github.com/MozillaSecurity/fuzzing-tc-config/pull/1

config/projects.yml
La0: Ok, i'll update that. I used `cd` to avoid using `os.path.expanduser`

config/projects.yml
La0: That avoids ssh warnings and using accept host keys. Usually i write this in the base docker image

config/projects.yml
La0: Done

PR #161, opened 2019-12-11
Add new hook scopes to bugbug deploy role

marco-c:
Would `assume:hook-id:project-relman/bugbug*` work? Or do I have to specify each?


marco-c:
> Star would work. Would you like to redraft to use that or should I land and apply as-is?

I've updated it.

djmitche:
applied

PR #166, opened 2019-12-11
Allow main bugbug hook to modify other bugbug hooks

marco-c:
- On release (tag), the main hook is deployed;
- When the main hook ends, the other hooks can be deployed (as they depend on the result of the main hook; if the main hook doesn't run or fails, the other hooks should be kept in their previous version).


djmitche:
applied

PR #167, opened 2019-12-12
Move aws constants in dedicated YAML file.

La0:
This file will be used by the [fuzzing team decision task](https://github.com/MozillaSecurity/fuzzing-tc) to build worker pools

I checked that this PR did *not* change anything when running `tc-admin diff` in comparison to current's master state.

### DIFF
--- a/generate/workers.py
+++ b/generate/workers.py
@@ -6,7 +6,8 @@
 
 from tcadmin.resources import WorkerPool, Secret, Role
 
-import copy, hashlib, json
+import copy, hashlib, json, os
+import yaml
 
 CLOUD_FUNCS = {}
 WORKER_IMPLEMENTATION_FUNCS = {}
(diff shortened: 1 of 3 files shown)


generate/workers.py
```diff
@@ -227,50 +228,12 @@ def aws(
...
-            "no-inbound": "sg-0659c2937ecbe7254",
-            "docker-worker": "sg-0f8a656368c567425",
-        },
-        "us-east-1": {
-            "no-inbound": "sg-07f7d21a488e192c6",
-            "docker-worker": "sg-08fea1235cf66b102",
-        },
-        "us-east-2": {
-            "no-inbound": "sg-00a9d64b3595c5088",
-            "docker-worker": "sg-0388de36e2f30ced2  u",
```
La0: I removed the trailing `u` here as it looks like a typo.

petemoore: I checked, and `sg-0388de36e2f30ced2` does exist in us-east-2, so that is correct now.

La0:
I moved the constants in `config/aws.yml` and added a description at the top of the file.

We are just starting to build the fuzzing decision task, so it should support edition and removal of values, but i left a comment about reaching out to the fuzzing team anyway...

djmitche:
Oops, yamllint failed.  Fix that up and this is good to merge.  Thanks!

PR #171, opened 2019-12-18
Add disable-seccomp scope to proj-taskcluster/ci for docker-worker

jschwartzentruber:
This adds the scope required by PR: https://github.com/taskcluster/docker-worker/pull/532 (disable seccomp)

I'm not sure if this needs to land before that PR or vice versa.

### DIFF
--- a/config/projects.yml
+++ b/config/projects.yml
@@ -93,6 +93,7 @@ taskcluster:
       workerConfig:
         dockerConfig:
           allowPrivileged: true
+          disableSeccomp: true
 
     windows2012r2-amd64-ci:
       owner: taskcluster-notifications+workers@mozilla.com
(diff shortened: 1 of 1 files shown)


config/projects.yml
```diff
@@ -93,6 +93,7 @@ taskcluster:
       workerConfig:
         dockerConfig:
           allowPrivileged: true
+          disableSeccomp: true
```
djmitche: Should we have a distinct worker to run docker-worker tasks, @walac?  This seems to be inviting trouble..

jschwartzentruber: On it's own, I believe this introduces less risk than `allowPrivileged` already does, although the combination is surely worse.

walac: This is necessary to test the new feature in docker-worker.

jschwartzentruber:
I removed the shm memory scope from this, since you also added it in #174.

config/projects.yml
djmitche: > This is necessary to test the new feature in docker-worker.

That makes sense.  I'm wondering if we should use a dedicated worker pool for docker-worker's own CI, and limit the capabilities on the `proj-taskcluster/ci` worker.

walac:
@jschwartzentruber would you mind rebasing your patch?

PR #174, opened 2019-12-19
Add new hostSharedMemory capability scope to docker-worker

walac:


jschwartzentruber:
This is for https://github.com/taskcluster/docker-worker/pull/533

PR #178, opened 2020-01-02
Bug 1598649 - add a tc 'release' worker

djmitche:


djmitche:
The release process uses `docker build` and a bunch of other docker commands that I don't trust to work via dind.  So basically the former.  I didn't need the osGroups property since the worker is running as the current user, which is already in the docker group.

PR #184, opened 2020-01-13
Bug 1607821 - *actually* shut down on idle

djmitche:


djmitche:
That seems like a good followup..

djmitche:
(applied, and all existing instance terminated)

PR #185, opened 2020-01-14
Add random-local-worker/docker-worker worker-id scopes

walac:


walac:
> I thought this got changed in the source to refer to a different workerPoolId?

Didn't work

PR #187, opened 2020-01-14
Add a new worker type specifically for bugbug regressor-finder task

marco-c:
This is exactly like the worker type above, but it is separate as I need a cache specifically for a specific task.


marco-c:
I'll use S3 actually, so this is not needed anymore.

PR #188, opened 2020-01-17
Bug 1588083 - use scheduler-id:smoketest for smoketests

djmitche:
Bugzilla Bug: [1588083](https://bugzilla.mozilla.org/show_bug.cgi?id=1588083)

Small change per @tomprince's request in https://phabricator.services.mozilla.com/D58272#inline-366030


djmitche:
(was already applied)

PR #190, opened 2020-01-23
Fix fuzzing hook scopes

jschwartzentruber:
This hook has been failing on missing scope `queue:create-task:medium:proj-fuzzing/decision` for some time.


djmitche:
(applied)

PR #191, opened 2020-01-23
Bug 1607935 - support static workers

djmitche:
This might be easiest to review commit-by-commit, as it was a bit of a refactor.

I only added one worker pool so far.  We can take that one to the finish line, then revisit for the others.

This *does* look forward to the day when we can provide config to static workers, at which time we can specify those in imagesets.


djmitche:
(rebased, waiting for CI..)

PR #193, opened 2020-01-26
split projects into separate files

djmitche:
The `projects.yml` file was getting a bit unweildy.  This should also make review a bit easier, where seeing that a change came from someone who "owns" the project in the filename is a pretty easy r+.


djmitche:
(I confirmed that `tc-admin diff` shows no differences after this change)

djmitche:
(re-confirmed that `tc-admin diff` shows the same diffs with this patch applied as it does on master)

PR #195, opened 2020-02-04
Install gcc on staging generic-worker worker types

petemoore:
This PR installs gcc on staging windows/ubuntu generic-worker worker types.

This is useful for overcoming [this issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1523934#c2) where gcc seems to be needed after upgrading to work with CGO. I'm not sure how it worked before without gcc.

Note, I commented out some parts of the staging ubuntu image set that installed generic-worker from source, rather than deleting them, since it is highly likely that this will be useful in future (typically we build from source on staging to get a development version before deciding whether to release or not).

Note, we can squash the commits if you are happy with the change.

Also note I've deployed the changes to the staging worker types already, so this PR just ensures the current source versions match what is currently deployed.

### DIFF
--- a/config/imagesets.yml
+++ b/config/imagesets.yml
@@ -74,11 +74,12 @@ generic-worker-win2012r2-staging:
   workerImplementation: generic-worker
   aws:
     amis:
-      us-east-1: ami-0162f92052875bf7e
-      us-west-1: ami-0f6b26130f2f9ce2c
-      us-west-2: ami-0dfd8045878e6cd08
-  gcp:
-    image: projects/community-tc-workers/global/images/generic-worker-win2012r2-staging-hasr18iepqtggvlswlku
+      us-east-1: ami-0f892d6ff7cdc52e7
+      us-west-1: ami-0cf711badfa2f50fc
+      us-west-2: ami-02ed2a69fdfdef0bc
+# Currently not deployed in gcp...
+# gcp:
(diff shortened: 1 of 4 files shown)


petemoore:
I think I'll close this and try another approach. Having investigated further, I really don't think we should need gcc at all, so I think this is an upstream issue. When installing gcc, the issue [isn't resolved](https://community-tc.services.mozilla.com/tasks/Nz4SsgHETVugRQ6EgcI5fw/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FNz4SsgHETVugRQ6EgcI5fw%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log#L2314-2320) and [this comment](https://github.com/golang/go/issues/12029#issuecomment-128060368) from Russ Cox 5 years ago suggests installing a [different toolchain](http://tdm-gcc.tdragon.net/) that hasn't been touched in the last 5 years. A bit of "stack overflowing" has led me to believe we shouldn't need gcc, since I suspect that is only needed when using gccgo, and we should be running cgo. So I'm going to instead try to understand why `go test` thinks it should be using `gccgo` instead of `gc`...

petemoore:
This is probably https://github.com/golang/go/issues/33840

petemoore:
Or possibly https://github.com/golang/go/issues/27089

PR #196, opened 2020-02-05
Remove fuzzing decision hook

La0:
As discussed with @petemoore & @jschwartzentruber  yesterday, we are replacing the fuzzing decision  hook by tasks triggered by the Github repo https://github.com/MozillaSecurity/fuzzing-tc-config


djmitche:
(applied)

PR #200, opened 2020-02-07
Increase idle timeout for generic-worker CI tasks

petemoore:
Originally attempted in #199 I realised afterwards I updated the wrong worker pool!

This time I was more careful. ;-)

Also updating the Windows Server 2012 R2 CI worker pool for similar gain.

Thanks!


djmitche:
There's a fuzzing change that can be applied, too, when you land this.

PR #201, opened 2020-02-11
Build fuzzing worker pools with Taskcluster CI/CD

La0:
This PR does several things:
1. Remove the useless `decision` worker pool, we only need `ci` to be static
2. Remove the privileged mode from `ci` worker pool, as the [docker image is now built with dind](https://github.com/MozillaSecurity/fuzzing-tc/issues/8)
3. As `tc-admin diff` now runs on every [MozillaSecurity/fuzzing-tc-config](https://github.com/MozillaSecurity/fuzzing-tc-config) pull request, we need to run `tc-admin apply` on every merge on master. So worker-manager scopes are requested for that branch.

### DIFF
--- a/config/projects/fuzzing.yml
+++ b/config/projects/fuzzing.yml
@@ -24,13 +24,6 @@ fuzzing:
       workerConfig:
         dockerConfig:
           allowPrivileged: true
-    decision:
-      owner: fuzzing+taskcluster@mozilla.com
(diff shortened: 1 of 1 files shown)


La0:
This blocks https://github.com/MozillaSecurity/fuzzing-tc-config/pull/6

djmitche:
.. & applied

PR #216, opened 2020-02-24
Add RTEMS project

### LINKED ISSUE #215: Add a new project for RTEMs
I've been chatting with @verm about Taskcluster, and it might be a good match for [RTEMS](https://www.rtems.org/).

The particular use-case involves bringing workers, so I think a good way to get started and see if there's a match is to set up a project in the community TC deployment, but not add any cloud-based workers.  The incremental cost to Mozilla is minimal, and if the experiment is successful then RTEMS can look into hosting their own Taskcluster deployment.

djmitche:
Fixes #215.


djmitche:
(already applied)

PR #217, opened 2020-02-25
Grant to Servo admins scope to notify their Matrix room

SimonSapin:
I don’t know if an alias like `#servo:mozilla.org` works of if the full ID `!crXxRMedJXFPxKffWI:mozilla.org` is required so this adds both so I can try them.

Does the dot in `mozilla.org` need to be escaped?

CC https://bugzilla.mozilla.org/show_bug.cgi?id=1522154

### DIFF
--- a/config/projects/servo.yml
+++ b/config/projects/servo.yml
@@ -26,6 +26,8 @@ servo:
     - grant:
         # notification
         - queue:route:notify.irc-channel.#servo.*
+        - queue:route:notify.matrix-room.#servo:mozilla.org.*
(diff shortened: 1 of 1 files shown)


djmitche:
Note that this support has not yet been released!  SOON! :)

SimonSapin:
Oh ok. I saw Matrix mentioned in https://docs.taskcluster.net/docs/reference/core/notify and wondered if it had been deployed :)

It’s only now copy-pasting that URL that I realize I was not looking it https://community-tc.services.mozilla.com/docs/reference/core/notify

PR #220, opened 2020-02-28
Bug 1596177: Grant non-legacy scopes for docker worker.

tomprince:


tomprince:
@djmitche If I am reading the code right, it looks the only thing that could be affected is https://github.com/taskcluster/docker-worker/blob/master/test/fixtures/create_images.js which uses the default scheduler id (which isn't granted to this repo). But it looks like that is run by hand, since https://github.com/taskcluster/docker-worker/blob/master/test/fixtures/image_artifacts.js points at a static task.

walac:
@tomprince yeah, you are right.

PR #221, opened 2020-02-28
Give fuzzing-tc-config repo access to deployment and credstash secrets for project/fuzzing

jschwartzentruber:


djmitche:
applied

jschwartzentruber:
Thanks!

PR #226, opened 2020-03-04
Add scopes to run tc-admin apply for proj-fuzzing

jschwartzentruber:
Also move some secret scopes to master branch only, where they are only needed for CD not CI.

### DIFF
--- a/config/projects/fuzzing.yml
+++ b/config/projects/fuzzing.yml
@@ -46,14 +46,20 @@ fuzzing:
         - repo:github.com/MozillaSecurity/fuzzing-tc:*
     - grant:
         - secrets:get:project/fuzzing/decision
-        - secrets:get:project/fuzzing/deploy-*
-        - secrets:get:project/fuzzing/credstash-aws-auth
       to:
         - repo:github.com/MozillaSecurity/fuzzing-tc-config:*
 
     # fuzzing-tc-config code on master can run tc-admin apply to manage worker pools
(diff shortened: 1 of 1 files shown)


La0:
LGTM :+1:

config/projects/fuzzing.yml
```diff
@@ -46,14 +46,20 @@ fuzzing:
...
+        - assume:hook-id:project-fuzzing/*
+        - auth:create-role:hook-id:project-fuzzing/*
+        - auth:update-role:hook-id:project-fuzzing/*
+        - hooks:modify-hook:project-fuzzing/*
+        - hooks:trigger-hook:project-fuzzing/*
+        - queue:scheduler-id:-
+        - secrets:get:project/fuzzing/credstash-aws-auth
+        - secrets:get:project/fuzzing/deploy-*
+        - worker-manager:manage-worker-pool:proj-fuzzing/*
+        - worker-manager:provider:community-tc-workers-*
```
petemoore: @imbstack is line 63 ok? I noticed the scope isn't specific to the fuzzing project, so I wanted to check with someone that knows more about worker manager. Thanks!

imbstack: That scope is required to create/update/delete worker pools that use those providers so if that's the intention to allow that here then it makes sense to me!

djmitche:
Applicated!

PR #234, opened 2020-03-13
Bug 1607935 - add support for lifecycle config in worker pools

djmitche:


config/projects/taskcluster.yml
```diff
@@ -97,6 +97,8 @@ taskcluster:
       emailOnError: true
       imageset: generic-worker
       cloud: static
+      lifecycle:
+        reregistrationTimeout: 2678400 # 31 days
```
imbstack: Won't the static provider reject this lifecycle config?

djmitche: [nope](https://docs.taskcluster.net/docs/reference/core/worker-manager/static) :)

That was added recently..

djmitche:
applied

PR #236, opened 2020-03-13
Add yamllint pre-commit hook

armenzg:


armenzg:
Once the user install the pre commit hooks with `pre-commit install` every commit will validate the user's staged files.

PR #239, opened 2020-03-16
Ds grants

lissyx:


config/imagesets.yml
```diff
@@ -170,3 +170,22 @@ deepspeech-win2012r2:
           machine-setup:
             maintainer: alissy@mozilla.com
             script: https://raw.githubusercontent.com/taskcluster/generic-worker/2509a1f8589b0b536186f5fe8f043385bf67197f/worker_types/deepspeech-win/bootstrap.ps1
```
lissyx: I don't know how we should update that, is it done manually?

config/imagesets.yml
```diff
@@ -170,3 +170,22 @@ deepspeech-win2012r2:
...
+    genericWorker:
+      config:
+        ed25519SigningKeyLocation: C:\generic-worker\generic-worker-ed25519-signing-key.key
+        shutdownMachineOnIdle: true
+        idleTimeoutSecs: 90
+        shutdownMachineOnInternalError: true
+        workerTypeMetadata:
+          machine-setup:
+            maintainer: alissy@mozilla.com
+            script: https://raw.githubusercontent.com/taskcluster/generic-worker/2509a1f8589b0b536186f5fe8f043385bf67197f/worker_types/deepspeech-win/bootstrap.ps1
```
lissyx: same here, this is to be synced

config/projects/deepspeech.yml
```diff
@@ -172,6 +172,7 @@ deepspeech:
         - repo:github.com/mozilla/tensorflow:*
     - grant:
         - generic-worker:allow-rdp:proj-deepspeech/*
+        - generic-worker:os-group:proj-deepspeech/*
```
lissyx: This is to be able to get RDP + admin rights if needed

imagesets/deepspeech-win2012r2-gpu/aws_filters
```diff
@@ -0,0 +1,2 @@
+Name=platform,Values=windows
+Name=name,Values=Windows_Server-2012-R2_RTM-English-64Bit-Base*
```
lissyx: Maybe there is a better AMI that already includes the Tesla M60 GPU drivers ?

config/imagesets.yml
petemoore: For now, yes - I can take care of that. In future this will be automatic, as it was in the past. Just in the interim it is manual until we have image building automatically as part of the CI of this project.

config/imagesets.yml
petemoore: 👍

config/projects/deepspeech.yml
petemoore: Fine. Be aware that any task will be able to run as an Administrator though, so make sure only trusted tasks can run.

imagesets/deepspeech-win2012r2-gpu/aws_filters
petemoore: I'm not sure, but I prefer it being done as you do here, so that the installation steps are explicit and under your own control. Some community images can lack transparency and also are not always very secure.

imagesets/deepspeech-win2012r2-gpu/gcp_filters
```diff
@@ -0,0 +1 @@
+--image=windows-server-2012-r2-dc-v20190910 --image-project=windows-cloud --boot-disk-size=50GB --boot-disk-type=pd-standard
```
petemoore: I'm not sure if we've tried running gpu instances in gcp before, so don't be surprised if there are issues that prevent it from working. But fine to add this for now so that we can try it out, and if something is broken, we can see what might need to be done to fix it.

PR #240, opened 2020-03-16
Add project Bergamot

djmitche:
Per email from Kelly Davis

### DIFF
--- a/config/projects/bergamot.yml
+++ b/config/projects/bergamot.yml
@@ -1,0 +1,25 @@
+bergamot:
+  adminRoles:
+    - github-team:mozilla/bergamot
+  repos:
+    - github.com/mozilla/marian-dev:*
+  workerPools:
+    ci-linux:
+      owner: nobody@mozilla.org
+      emailOnError: false
+      imageset: docker-worker
+      cloud: gcp
+      minCapacity: 0
+      maxCapacity: 10
+    ci-windows:
+      owner: nobody@mozilla.org
+      emailOnError: false
+      imageset: generic-worker-win2012r2
+      cloud: aws
+      minCapacity: 1
+      maxCapacity: 10
+  grants:
+    - grant:
+        - queue:create-task:highest:proj-bergamot/ci-*
+      to:
+        - repo:github.com/mozilla/marian-dev:*


djmitche:
@kdavis-mozilla how does this look?  This will give you a dedicated worker pool, `project-bergamot/ci`, running docker-worker in CI.  There's one TODO in there, too.  I can list you as admin, but maybe there's already a GH team in place we could use?

You'll also need to ask an admin of the @mozilla GH org to add the `marian-dev` repository to the [GitHub App](https://github.com/apps/community-tc-integration).

config/projects/bergamot.yml
```diff
@@ -0,0 +1,17 @@
+bergamot:
+  adminRoles: [] # TODO github teams or users who should have admin access to this project
```
kdavis-mozilla: I just created a team for this [github-team:mozilla/bergamot](https://github.com/orgs/mozilla/teams/bergamot) which right now only has myself.

config/projects/bergamot.yml
```diff
@@ -0,0 +1,17 @@
...
+  repos:
+    - github.com/mozilla/marian-dev:*
+  workerPools:
```
kdavis-mozilla: I'd need linux, windows, and mac workers as I need to make sure the code continues to work there too.

config/projects/bergamot.yml
```diff
@@ -0,0 +1,17 @@
...
+    ci:
+      owner: nobody@mozilla.org
+      emailOnError: false
+      imageset: docker-worker
+      cloud: gcp
+      minCapacity: 0
+      maxCapacity: 10
```
kdavis-mozilla: Would I need clients for each of the various worker types, e.g. a windows client?

config/projects/bergamot.yml
djmitche: We can add a Windows cloud-based worker-pool, too.  However, nobody offers macs in pools, so for that platform we'll need to find a mac (or more than one).  That might be in MacStadium, or another mac-rental company, or a mac in a Mozilla office somewhere.  That might be a good next step after we get the rest started.

config/projects/bergamot.yml
djmitche: Sounds good!

config/projects/bergamot.yml
djmitche: For the most part, you just create a worker pool for each.  For non-cloud things (like macs), currently that requires creating a client, but we're working on some final blockers to making that easy, too.  So we'll see how the timing works out.

djmitche:
Updated.  The worker pool name changed, to include context.  It's now `project-bergamot/ci-linux` and `project-bergamot/ci-windows`.

config/projects/bergamot.yml
kdavis-mozilla: Thanks on the windows workers.

And on the Mac workers, yeah the Deep Speech mac sits one the desk next to me. I hear every PR :)

djmitche:
applied (I think.. lots of grepping around going on)

kdavis-mozilla:
@djmitche One question, which may be stupid as I'm learning TC...

[bergamot.yml](https://github.com/mozilla/community-tc-config/blob/master/config/projects/bergamot.yml) specifies a ci-linux worker pool. But looking at the [page](https://community-tc.services.mozilla.com/provisioners/proj-bergamot) showing workers for proj-bergamot there are no ci-linux workers only ci-windows workers. Is this correct?

djmitche:
It's not a stuipd question -- the UI is misleading :)  The worker page you're looking at shows only workers that have actually run.  So I'm guessing that there haven't been any ci-linux workers lately, and thus the worker pool does not appear there.  Both worker pools do appear [here](https://community-tc.services.mozilla.com/worker-manager).  There's work currently underway to make this better!

kdavis-mozilla:
Thanks!

PR #244, opened 2020-03-17
Grant scopes to CIA hook and repo

armenzg:
Is it fine that I grant `secrets:get:project/cia/garbage/foo` to both the hook and the repo?


djmitche:
applied

PR #249, opened 2020-03-19
Added grants required of decision task

kdavis-mozilla:
I've added grants required of [decision task](https://github.com/mozilla/marian-dev/blob/master/.taskcluster.yml)


djmitche:
applied

PR #252, opened 2020-04-01
Grant to Servo admins RDP access to Servo workers

SimonSapin:


SimonSapin:
@petemoore This diff is based on the precedent of https://github.com/mozilla/community-tc-config/blob/9e3918d97aa629a300c403a84c20da49a3c7b1d4/config/projects/deepspeech.yml#L174, but would it make sense to grant `generic-worker:allow-rdp:proj-<..>/*` to `project-admin:*` instead?

imbstack:
applied!

PR #254, opened 2020-04-24
Update community image to drop invalid packets for wpt#21529

imbstack:
I notice that the last image used here was not the `community` variant. Should we always be using the community variant here?

### DIFF
--- a/config/imagesets.yml
+++ b/config/imagesets.yml
@@ -48,10 +48,16 @@ generic-worker:
   workerImplementation: generic-worker
   # (this is intended for use with static workers)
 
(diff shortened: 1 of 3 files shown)


imbstack:
Another option here is to just roll it out for wpt and taskcluster so that in case it breaks things we don't break things over the weekend for projects we're not watching closely

ccooper:
We should be making the distinction between the resources in different clusters, yes. We've had  issues in the past -- security and otherwise -- where resources got mixed up due to inadequate nomenclature. Even if the resources are functionally identical right now, that might not always be the case.

But yes, let's not fix that issue late on a Friday afternoon.

PR #261, opened 2020-05-01
Bug 1634589 - custom AWS instance type per image set for base images

petemoore:
Long term we'd like to move to monopacker, but this is a stopgap that allows specifying a custom AWS instance type for building an image set.

Previously, all image sets were _built_ using `c4.2xlarge`, but then when instances were launched by Worker Manager, it would use the instance type specified in the worker pool definition. This was fine most of the time, we hit an issue though when gpu capabilities were required during image building, in deepspeech gpu windows worker types.

Therefore, this change removes the hardcoded `c4.2xlarge` reference in `imageset.sh` and instead fetches the value from a configuration file (`aws_base_instance_type`) inside the image set directory.

It also takes care to ensure that if a particular instance type isn't supported in a region, to report it in the output, so that worker deployers understand why not all regions have AMIs generated.

See [bug 1634589](https://bugzil.la/1634589).


petemoore:
Thanks!

PR #263, opened 2020-05-08
Add a 'compute-smaller' worker type which is bigger than 'batch' and smaller than 'compute-small'

marco-c:


djmitche:
```
Creating WorkerPool=proj-relman/compute-smaller
Creating Role=worker-pool:proj-relman/compute-smaller
Creating Secret=worker-pool:proj-relman/compute-smaller
```

PR #265, opened 2020-05-11
Bug 1636174 - enable docker-worker CI in taskcluster monorepo PRs

petemoore:
Let's enable docker-worker CI on taskcluster monorepo PRs.

See [bug 1636174](https://bugzil.la/1636174) for context.


djmitche:
#266 does one-better, with a distinct worker for these jobs

PR #268, opened 2020-05-13
Add web-platform-tests/wpt-tc-checks to the checks/statuses workaround

stephenmcgruer:
WPT is moving towards using the checks API for TaskCluster. However we
also use a decision task and as per https://github.com/mozilla/community-tc-config/pull/129
this requires us to be added to the route workaround.

Note that wpt-tc-checks is a temporary clone of wpt being used as we
develop the TaskCluster Checks integration; ultimately this will be
replaced with the wpt repo itself.

### DIFF
--- a/config/grants.yml
+++ b/config/grants.yml
@@ -186,6 +186,7 @@
   to:
     # Repositories that whish to use this work-around
     - repo:github.com/servo/servo:*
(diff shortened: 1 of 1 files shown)


djmitche:
applied

PR #269, opened 2020-05-13
Grant wpt-tc-checks the same scopes as wpt

stephenmcgruer:
The wpt-tc-checks repo is a temporary* clone of wpt, used to develop our
necessary integration to switch from TaskCluster statuses to checks. To
let TaskCluster run on it, it needs the same grants as wpt.

* Temporary here means O(weeks); the hardest part of this project will
be fixing wpt.fyi to work with the checks API and I expect that to take
me a few weeks.


djmitche:
applied

PR #270, opened 2020-05-14
add S3 permission to cia hooks

klahnakoski:


djmitche:
@marco-c can I get your a-ok for this?

djmitche:
applied

PR #271, opened 2020-05-15
Shutdown proj-relman/compute-large and proj-relman/compute-super-large right away

marco-c:


djmitche:
(sandbox) rubin ~/p/community-tc-config [master] $ tc-admin apply  --ids-only
Updating WorkerPool=proj-relman/compute-super-large
Updating WorkerPool=proj-relman/compute-large

PR #273, opened 2020-05-15
Add create-task scope to fuzzing-tc-config repo.

jschwartzentruber:


djmitche:
```
(sandbox) rubin ~/p/community-tc-config [master] $ tc-admin apply  --ids-only --without-secrets
Updating Role=repo:github.com/MozillaSecurity/fuzzing-tc-config:branch:master
```

PR #278, opened 2020-05-26
Reduce minCapacity to 0 everywhere

ccooper:
This will have the side-effect of slightly longer run times as we wait
for workers to be provisioned, but looking at the costs for time spent
idle, it would be hard to justify *not* making this change.

I've also explicitly added the minCapacity in some places where it was
missing (e.g. projects/taskcluster) instead of relying on the default
behavior. This will make it clearer when we look back when the change
was made.

While not stricly necessary, I've also reduced the maxCapacity for many
worker pools. The volume of jobs just doesn't justify the previous
requested max values.

### DIFF
--- a/config/projects/relman.yml
+++ b/config/projects/relman.yml
@@ -27,8 +27,8 @@ relman:
       emailOnError: false
       imageset: docker-worker
       cloud: gcp
-      minCapacity: 1
-      maxCapacity: 50
+      minCapacity: 0
+      maxCapacity: 10
       workerConfig:
         dockerConfig:
           allowPrivileged: true
@@ -37,32 +37,32 @@ relman:
       emailOnError: false
       imageset: docker-worker
       cloud: gcp
-      minCapacity: 1
-      maxCapacity: 25
+      minCapacity: 0
+      maxCapacity: 10
       machineType: "zones/{zone}/machineTypes/n1-standard-1"
     compute-smaller:
       owner: mcastelluccio@mozilla.com
       emailOnError: false
       imageset: docker-worker
       cloud: gcp
-      minCapacity: 1
-      maxCapacity: 25
+      minCapacity: 0
+      maxCapacity: 10
       machineType: "zones/{zone}/machineTypes/n1-standard-2"
     compute-small:
       owner: mcastelluccio@mozilla.com
       emailOnError: false
       imageset: docker-worker
       cloud: gcp
-      minCapacity: 1
-      maxCapacity: 25
+      minCapacity: 0
+      maxCapacity: 10
       machineType: "zones/{zone}/machineTypes/n1-standard-4"
     compute-large:
       owner: mcastelluccio@mozilla.com
       emailOnError: false
       imageset: docker-worker
       cloud: gcp
-      minCapacity: 1
-      maxCapacity: 25
+      minCapacity: 0
+      maxCapacity: 10
       machineType: "zones/{zone}/machineTypes/n1-standard-8"
       workerConfig:
         genericWorker:
@@ -74,8 +74,8 @@ relman:
       emailOnError: false
       imageset: docker-worker
       cloud: gcp
-      minCapacity: 1
-      maxCapacity: 25
+      minCapacity: 0
+      maxCapacity: 10
       machineType: "zones/{zone}/machineTypes/n1-standard-16"
       workerConfig:
         genericWorker:
@@ -87,6 +87,7 @@ relman:
       emailOnError: false
       imageset: deepspeech-win2012r2
       cloud: aws
+      minCapacity: 0
       maxCapacity: 10
   secrets:
     bugbug/deploy: true
--- a/config/projects/relman.yml
+++ b/config/projects/relman.yml
@@ -28,7 +28,7 @@ relman:
       imageset: docker-worker
       cloud: gcp
       minCapacity: 0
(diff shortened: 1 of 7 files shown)


ccooper:
> I agree it's hard to see how this isn't a good idea, on a cost basis. Let's let the various involved folks know, though. A bit of `git log` and looking at some PRs should identify the necessary folks.

Good call. I'll compile the list and let them know prior to merging.

config/projects/relman.yml
```diff
@@ -37,32 +37,32 @@ relman:
       emailOnError: false
       imageset: docker-worker
       cloud: gcp
-      minCapacity: 1
-      maxCapacity: 25
+      minCapacity: 0
+      maxCapacity: 10
```
marco-c, line 41: If `maxCapacity` means the maximum number of workers that can run in parallel, some of the values for the `relman` workers could be lower (if it's worth it).

ccooper: The minCapacity is more important than the maxCapacity in terms of carrying costs. Given the volume going through the system, that single instance running forever is much more expensive than 10 instances claiming tasks once a week and then shutting down.

ccooper:
@kdavis-mozilla @armenzg @SimonSapin @stephenmcgruer Wanted to flag this change to you. The volume of jobs going through the community cluster is very small, so having standing worker pools, even small ones, is hard to justify.

stephenmcgruer:
Thanks for the heads up @ccooper (cc @jgraham as well). 

> slightly longer run times as we wait for workers to be provisioned

Do you have numbers on how much longer, on average? I'm mostly trying to get a sense how this will affect latency for the CI on pull requests, e.g. the time between someone opening a pull requests and then having Taskcluster start going.

config/projects/wpt.yml
```diff
@@ -5,7 +5,8 @@ wpt:
     ci:
       imageset: docker-worker
       cloud: gcp
-      maxCapacity: 80
+      minCapacity: 0
+      maxCapacity: 10
```
stephenmcgruer, line 9: An 8-fold decrease is quite a lot - were we just never near this capacity? That seems a little strange to me since I have seen Taskcluster jobs occasionally that seem to be waiting for 'something' to start, and I always assumed it was CI resources. cc @jgraham

jgraham: When started on the community instance we had to bump this number; see https://github.com/mozilla/community-tc-config/commit/1a844c921b3dc13120e8a54a654061de2dd0468c

10 seems likely too low to continue our current usage pattern; if this is a requirement I think we're going to have to look for either changes to the usage or changes to the CI system we use.

djmitche: If I remember, WPT's usage is pretty spiky, with steep spikes -- 100's of tasks created at once.  So Scaling up to run all or most of those tasks in parallel has a big impact on E2E time.

SimonSapin:
Do we have numbers of the actual respective usage of workers in each worker pool? Are they public?

Servo’s CI is rather sensitive to latency: Homu a.k.a. Bors is based on the idea of only pushing to `master` exactly the merge commit that has been tested, after tests pass. Without doing speculation or batching, this means effectively means testing one PR at a time. Testing one PR takes about an hour and involves dozens of tasks. So the effects of increasing latency add up. Sometimes the queue https://build.servo.org/homu/queue/servo is empty, but sometimes it looks more like this:



I don’t think I’ve measured it precisely, but my recollection is that starting a new VM on GCP adds a couple minutes of latency (for a single task), compared to having a worker already running and waiting for tasks which is a lot more likely with non-zero `minCapacity`.

Testing PRs for landing is serialized, but "try" testing can be parallel and uses the same worker pools. If workers are limited, having a couple "try" PRs at the same time can starve the main queue. Testing a single PR takes 13 tasks on Linux/GCP at the moment, so `maxCapacity: 10` seems like a steep limitation.

I’m very willing to work on reducing CI cost, but it’d help to see where the costs are in the first place. I think there’s a number of things we can do, like using lower-specs VMs for tasks like decision tasks that don’t need a high CPU core count. Or tweak the time a worker stays idle before it shuts down.

config/projects/wpt.yml
SimonSapin: With spiky usage, it seems that reducing idle-time-until-shutdown for each worker would be more productive than reducing `maxCapacity`.

marco-c:
> I think there’s a number of things we can do, like using lower-specs VMs for tasks like decision tasks that don’t need a high CPU core count

I wish there was a easy way to monitor resource usage and figure out what worker pools are overprovisioned in terms of CPU/memory (same with the capacities).
I have a feeling many of the worker pools are used for very simple tasks and are needlessly n1-standard-4 (https://bugzilla.mozilla.org/show_bug.cgi?id=1636455).

ccooper:
I'm going to hold off on merging this for a bit while I do some further investigation. I don't have a lot of insight into the GCP data, and since most of the community cluster is running in GCP, I want to validate some assumptions before hamstringing everyone's CI unnecessarily.

ccooper:
I reverted the maxCapacity changes. I paid particular attention to WPT last night, and instances were cycling as expected.  AFAICT we're cleaning up instances properly, but the GCP data is too coarse to tell where the overhead is coming from, so I'll leave the maxCapacity alone so project can continue to have their burst capacity until we learn more.

PR #286, opened 2020-06-03
Create a client for running community worker version reports

petemoore:
This client is used for running [worker version reports](https://github.com/taskcluster/community-history/tree/master/WorkerVersions/README.md) against our cluster.

It works by submitting tasks for all discovered (provisioner, worker type) pairs, excluding those known to be dynamically generated for CI tests, and then parsing the artifacts of the generated tasks.

Note, we can probably replace this eventually with an exchange of metadata when workers register with Worker Manager. However, for now this is a pragmatic approach to obtaining the data, and has the advantage that it also works for workers that do not interface with worker manager (such as manually provisioned workers that might not be yet using the static provider type). It allows you to detect workers attached to your cluster that you might otherwise be unaware of.

I'm working on an updated version that doesn't need the credentials, and can parse logs from existing tasks. This has the advantages of being faster (no need to wait for task to be claimed) and requiring no scopes, however it has the disadvantages that there may not be recent tasks for all worker types (so not all worker type versions can be inferred) and also that the data is less current (updated worker pools that haven't run tasks since they were updated will report the version they had when the most recent task ran) and also that it doesn't help you promptly/routinely find badly configured workers. In contrast, when routinely submitting tasks
for the worker report, you are essentially reporting on worker pool health. You see straight away if you have worker pools that are not able to claim tasks, because the task you created isn't getting claimed (e.g. large pending queues, misconfiguration of worker pool config, broken hardware, crashed machines, .....).

Therefore the approach of submitting tasks has many favourable side effects, so on the whole I prefer it, when there isn't a security concern about running this task on any worker.

### DIFF
--- a/config/projects/community-admin.yml
+++ b/config/projects/community-admin.yml
@@ -1,0 +1,6 @@
+community-admin:
+  clients:
+    audit-worker-versions:
+      scopes:
+        - queue:create-task:highest:*
+      description: "Client used for [worker version](https://github.com/taskcluster/community-history/tree/master/WorkerVersions) reports."
--- a/config/projects/community-admin.yml
+++ b/config/projects/community-admin.yml
@@ -2,5 +2,5 @@ community-admin:
   clients:
     audit-worker-versions:
       scopes:
-        - queue:create-task:highest:*
+        - queue:create-task:lowest:*
       description: "Client used for [worker version](https://github.com/taskcluster/community-history/tree/master/WorkerVersions) reports."


imbstack:
Should we let all  projects know before we start doing something like this? Also should this use `lowest` priority? I guess I don't have any huge concerns about this scope but we do have people with special workers they care about in this deployment I think.

petemoore:
> Should we let all projects know before we start doing something like this? Also should this use `lowest` priority? I guess I don't have any huge concerns about this scope but we do have people with special workers they care about in this deployment I think.

That's fair. I've switched the tasks to run with `lowest` priority. Note, the task payloads are intentionally invalid, so that workers can quickly resolve them as `exception/malformed-payload` with minimal overhead.

@imbstack Do we have a distribution list of existing contact list for contacting all projects?

ccooper:
> Should we let all projects know before we start doing something like this? Also should this use `lowest` priority? I guess I don't have any huge concerns about this scope but we do have people with special workers they care about in this deployment I think.

My concern is about cadence. Do worker versions drift so frequently that we need a daily report? I say this from the standpoint of someone who has spent the last few weeks digging into community costs and overhead. Each one of these tasks represents extra load and hence expense and I don't think the output changes that often.

I would be fine with a report we could trigger when needed, possibly only when we do a release.

ccooper:
> Note, we can probably replace this eventually with an exchange of metadata when workers register with Worker Manager.

This is what I want. We have (or can easily acquire) this data via the worker-manager, so let's do that as the better, long-term solution.

@petemoore I don't anticipate that this is a quick fix, so can we shelve this for now?

petemoore:
Sure, we can shelve it.

PR #293, opened 2020-06-12
upgrade docker-worker to 30.1.1

djmitche:
Seems to be working in testing..

https://community-tc.services.mozilla.com/tasks/QrjKgfgmSBKWxJHghUZcjw


djmitche:
applied

PR #294, opened 2020-06-14
Use preemptible GCP instances.

tomprince:
I noticed that the GCP configuration wasn't using preemptible instances. I wasn't sure if this was deliberate or an oversight.


imbstack:
I'll try to come up with a metric and land this later today or this week.

imbstack:
And by "this week", I meant "next week" obviously :smile: 

Reading [the preemption docs](https://cloud.google.com/compute/docs/instances/preemptible) I see that in gcp preemptible instances can live a maximum of 24 hours so I'm thinking we should also configure the maximum lifespan of a worker to be 24 hours here as well? I doubt we hit that very often in community but could be nice to do up front.

PR #295, opened 2020-06-15
Add web-platform-tests/wpt to the checks/statuses workaround

stephenmcgruer:
WPT is moving towards using the checks API for TaskCluster. However we
also use a decision task and as per #129
this requires us to be added to the route workaround.

Note that wpt-tc-checks is a temporary clone of wpt being used as we
develop the TaskCluster Checks integration; it will be removed once
we've confirmed that the migration works.


djmitche:
applied

PR #296, opened 2020-06-16
Allow WPT to use built-in/success

stephenmcgruer:
WPT intends to add a sink task to their dependency graph, as part of
their migration to the GitHub Checks API. As this task is a do-nothing,
it should use built-in/success, but this requires extended grants.


tomprince:
I suspect all projects should be granted access to `built-in/*` workers.

djmitche:
applied!

PR #297, opened 2020-06-16
Correct typo in grant for WPT

stephenmcgruer:
It's called `build-in/succeed`, not `built-in/success`, but in either case giving access
to all of built-in seems reasonable.

See https://docs.taskcluster.net/docs/reference/workers/built-in-workers


djmitche:
applied

PR #300, opened 2020-06-25
Update firefoxreality.yml

owlishDeveloper:
FxR will be moving to the `main` branch being their default as opposed to `master`. We are adding the index scope to the new role


djmitche:
I seem to have messed up my secrets access so I can't land this right now -- can someone else on the team do so?

ccooper:
This has been applied now too.

PR #309, opened 2020-07-09
[fuzzing] Clean-up and grant cancel-task to decision.

jschwartzentruber:
* Remove the list of secrets since it's incomplete anyways and we're externally managed.
* Remove podman pool used for experiment.
* Grant `queue:cancel-task:-/*` to decision so we can cancel old tasks.


djmitche:
applied

PR #311, opened 2020-07-12
Change UWP requirements to visualstudio2019buildtools

carlfm01:
@lissyx Who should I ping for a review?


lissyx:
That would be @pmoore

lissyx:
> So we can roll it out as you have it, but that may not be what you want. Can you confirm?

This is actually what we want, we need to take the time to fix this in the future, but we have not yet had the luxury to do so, and such current status is this mixed state.

petemoore:
> > So we can roll it out as you have it, but that may not be what you want. Can you confirm?
> 
> This is actually what we want, we need to take the time to fix this in the future, but we have not yet had the luxury to do so, and such current status is this mixed state.

OK no worries. Deploying now!

PR #312, opened 2020-07-14
Check region secret exists prior to using

petemoore:
If a given aws region does not support the instance type required for a particular worker type, it is skipped. However, the `imageset.sh` script previously did not check if it was skipped before updating `config/imagesets.yml` and the password store secrets for a given (imageset, region).

Now `imageset.sh` no longer assumes an image is present in a region, and checks first, before applying changes that require the image to exist.


petemoore:
Thanks!

PR #316, opened 2020-07-24
Switch deepspeech ci worker to n2d-highcpu-32

lissyx:
According to GCP docs:
 - n2d is faster, newer gen
 - highcpu is more balanced between CPU / RAM
 - we don't need the 120GB RAM of n1-standard-32
 - it's cheaper (price 1.52$ -> 0.998$ ; preemptive price 0.32$ -> 0.2415$)

### DIFF
--- a/config/projects/deepspeech.yml
+++ b/config/projects/deepspeech.yml
@@ -12,7 +12,7 @@ deepspeech:
       cloud: gcp
       minCapacity: 0
       maxCapacity: 96
-      machineType: "zones/{zone}/machineTypes/n1-standard-32"
+      machineType: "zones/{zone}/machineTypes/n2d-highcpu-32"
     win:
       owner: deepspeech@mozilla.com
       emailOnError: false


lissyx:
There is one question, though: it is mentionned that nested virtualization is not available on n2d. We are using KVM-exposed-through Docker for running some virtualization-enabled kvm/qemu code (Android Emulator), but I'm unsure if that makes us using nested virtualization (i.e., is docker-worker running in a VM already or not?).

djmitche:
@ccooper does this sound good from a cost standpoint?

lissyx:
Sorry, looks like GCP is full of traps:
 - we're hitting the nested virtualization case, so we can't just switch like that, we need more work on our taskcluster setup to decouple tasks requiring nested virtualization VS those dont
 - n2d is not available everywhere, especially only us-central1 in the US, so we'd need more changes than that.

PR #317, opened 2020-07-24
[Cleanup] Remove old wpt-tc-checks repository

stephenmcgruer:
wpt-tc-checks was a testing-ground repository for wpt as it moved to the
Taskcluster Checks integration. That work has now been completed, and
the repository is no longer needed.


djmitche:
applied

PR #318, opened 2020-07-27
Use n2-highcpu-32

lissyx:
This machine type:
 - allows for nested virtualization
 - is cheaper (1.5$ -> 1.1471$ ; 0.32$ -> 0.2776$)
 - newer and should be faster
 - is available in us-east1 and us-east4 zones

So I think it should be an easier switch for the moment than moving to n2d-highcpu-32

### DIFF
--- a/config/projects/deepspeech.yml
+++ b/config/projects/deepspeech.yml
@@ -12,7 +12,7 @@ deepspeech:
       cloud: gcp
       minCapacity: 0
       maxCapacity: 96
-      machineType: "zones/{zone}/machineTypes/n1-standard-32"
+      machineType: "zones/{zone}/machineTypes/n2-highcpu-32"
     win:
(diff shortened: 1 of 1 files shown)


djmitche:
applied

reuben:
We're getting errors after this change:

Quota 'N2_CPUS' exceeded. Limit: 24.0 in region us-east1.	
Quota 'N2_CPUS' exceeded. Limit: 24.0 in region us-east4.

And:
Invalid value for field 'resource.machineType': 'zones/us-east1-b/machineTypes/n2-highcpu-32'. Machine type with name 'n2-highcpu-32' does not exist in zone 'us-east1-b'.

https://community-tc.services.mozilla.com/worker-manager/proj-deepspeech%2Fci/errors

PR #320, opened 2020-07-27
Revert "Use n2-highcpu-32"

reuben:
Reverts mozilla/community-tc-config#318 due to AWS errors so we can get tasks going again.


reuben:
Closing, looks like there's a quick fix.

PR #322, opened 2020-07-28
copy perms from tc monorepo master to main

djmitche:
See https://github.com/taskcluster/taskcluster/issues/3276


djmitche:
(this is already applied)

PR #325, opened 2020-07-29
Create a separate configuration for bugbug, with admin from both sallt and cia teams

marco-c:
As suggested by @djmitche in #319, I'll split the patch in two: one to create the new bugbug project (to be landed right away) and one to remove bugbug from the relman project (to be landed after bugbug is updated to use the worker pool names from the new project).


djmitche:
applied

PR #326, opened 2020-07-29
Grant create-task for proj-bugbug instead of proj-relman in the bugbug project

marco-c:
The other things (hook IDs, secret names, etc.) will keep "relman" in their name, so we don't need to change too many things.


marco-c:
Sorry, wrongly closed this as I thought it had been merged (I was confused by GitHub emails for #319 :P)

djmitche:
applied

PR #327, opened 2020-08-01
Update proj-deepspeech/ci workers to 100GB disk

lissyx:
This is hopefully going to help avoid tensorflow linux cuda build failure like:
> Server terminated abruptly (error code: 14, error message: 'Socket closed', log file: '/home/build-user/.bazel_cache/output/server/jvm.out')

### DIFF
--- a/config/projects/deepspeech.yml
+++ b/config/projects/deepspeech.yml
@@ -13,6 +13,7 @@ deepspeech:
       minCapacity: 0
       maxCapacity: 96
       machineType: "zones/{zone}/machineTypes/n2-highcpu-32"
+      diskSizeGb: 100
     win:
       owner: deepspeech@mozilla.com
       emailOnError: false


djmitche:
Do you think this will resolve https://bugzilla.mozilla.org/show_bug.cgi?id=1652998?

djmitche:
applied

lissyx:
I'd love it solves this bug, but i doubt.

lissyx:
Thanks for the quick review, i guess we'll have that deployed monday ? If so, that'd be perfect 😊

djmitche:
it's deployed now

lissyx:
> it's deployed now

Thanks. I don't know if this was being used by this re-run #16: https://community-tc.services.mozilla.com/tasks/SomTmnCWSASyY4eulHmU4w/runs/16

If yes, it means it was not our issue, and the inability to access any log is going to make that hard to debug. I fear we might need to revert to `n1-standard-32` soonish ...

djmitche:
It would only affect newly-created workers, so if that worker existed before this was landed, then it's not a 100G worker.

lissyx:
> It would only affect newly-created workers, so if that worker existed before this was landed, then it's not a 100G worker.

Right, I'm making a re-trigger (rerun can't work now since the deadline is expired): https://community-tc.services.mozilla.com/tasks/TqIAeYh4QDG-8vpSE0v-NQ to verify, since there is no worker running now.

And if that's still no good, I already have sent https://github.com/mozilla/community-tc-config/pull/330 :/

lissyx:
> > It would only affect newly-created workers, so if that worker existed before this was landed, then it's not a 100G worker.
> 
> Right, I'm making a re-trigger (rerun can't work now since the deadline is expired): https://community-tc.services.mozilla.com/tasks/TqIAeYh4QDG-8vpSE0v-NQ to verify, since there is no worker running now.
> 
> And if that's still no good, I already have sent #330 :/

Yeah, confirmed it's still not working:
> Server terminated abruptly (error code: 14, error message: 'Socket closed', log file: '/home/build-user/.bazel_cache/output/server/jvm.out')

PR #330, opened 2020-08-02
Revert to n1-standard-32 for linux/cuda build breakage

lissyx:


lissyx:
@djmitche I'm really sorry for the back/forth on that, so I'm reverting to n1-standard-32 because I fear we are exhausting the 32GB RAM on the tensorflow linux cuda builds, and I'm introducing finer-grained workers definitions so I can try without breaking stuff.

lissyx:
One last try on n2-highcpu-32, just in case it was 100GB disk space not yet there (but I'm not really convinced now): https://community-tc.services.mozilla.com/tasks/TqIAeYh4QDG-8vpSE0v-NQ

lissyx:
> One last try on n2-highcpu-32, just in case it was 100GB disk space not yet there (but I'm not really convinced now): https://community-tc.services.mozilla.com/tasks/TqIAeYh4QDG-8vpSE0v-NQ

So we can go ahead with reverting to n1, it's still failing:
> Server terminated abruptly (error code: 14, error message: 'Socket closed', log file: '/home/build-user/.bazel_cache/output/server/jvm.out')

PR #331, opened 2020-08-03
Win workers sync

lissyx:


djmitche:
This is built and images deployed.

PR #332, opened 2020-08-03
don't expect shell builtins to exist on the fs

djmitche:


djmitche:
For example, `cd` is not found on my system.

PR #333, opened 2020-08-03
Add per-worker device/privileged scopes for docker-worker tests.

tomprince:
This is required for the new test in taskcluster/taskcluster#3326


djmitche:
applied

PR #334, opened 2020-08-03
Remove bugbug-related repos, worker pools, etc. from relman project configuration

marco-c:


marco-c:
> Looks fine, what happened to bugbug? Has it been replaced by something else?

I moved it to a separate config file (config/projects/bugbug.yml), to grant admin rights to both mozilla/sallt and mozilla/cia.

PR #338, opened 2020-08-26
Simplify and fix for worker-runner deepspeech macOS

lissyx:


petemoore (approved):
Looks good to me, many thanks. I'll deploy after merging.

PR #342, opened 2020-08-27
Clean up scopes for master->main renames

djmitche:
json-e and monopacker have had their branches renamed, but the scopes
weren't updated.  Further, we just dropped support for doing anything
special on pushes to monorepo's main branch.  There's no longer a need
for the old "master" scope for the monorepo, either.


petemoore (approved):
Looks good, thanks!

djmitche:
applied

PR #343, opened 2020-08-28
Rename all bugbug secrets and hooks by removing the 'relman' string and set back externallyManaged to true for relman

marco-c:


djmitche:
applied

PR #346, opened 2020-09-09
Add create-task scope for bugmon-monitor

pyoor:
Fixes incorrect create-task scoping for bugmon-monitor.


petemoore (approved):
Looks good, many thanks!

PR #350, opened 2020-09-16
Bug 1665052 - use updated docker-worker images

djmitche:
These are the images built in
https://github.com/taskcluster/monopacker/pull/85

### DIFF
--- a/config/imagesets.yml
+++ b/config/imagesets.yml
@@ -51,13 +51,13 @@ docker-worker:
   workerImplementation: docker-worker
   gcp:
     # built with the `docker_worker_gcp_community` builder in monopacker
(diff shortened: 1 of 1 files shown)


djmitche:
https://github.com/taskcluster/taskcluster/issues/3521

djmitche:
Comments in https://bugzilla.mozilla.org/show_bug.cgi?id=1665052 but these latest images are:
 - d-w 30.1.1
 - confirmed to contain the latest monopacker
 - successfully run test tasks with loopback audio/video and a decision task to test for taskcluster/taskcluster#3521

petemoore (approved):
The scope of the change looks correct, and since your sample tasks ran successfully, it is likely that the image names are also correct (at least the ones the test tasks used). Many thanks!

djmitche:
applied

PR #354, opened 2020-10-23
Add anonymous role

helfi92:
I'm not sure with `to: *`. Please advise on the proper way to grant to everybody.

### DIFF
--- a/config/grants.yml
+++ b/config/grants.yml
@@ -196,3 +196,8 @@
 # Allow the taskcluster team to handle the denylist
 - grant: notify:manage-denylist
   to: github-team:taskcluster/core
+
+# Make all endpoints public by default
+- grant: assume:anonymous
+  to: *
+
--- a/config/grants.yml
+++ b/config/grants.yml
@@ -197,7 +197,7 @@
 - grant: notify:manage-denylist
   to: github-team:taskcluster/core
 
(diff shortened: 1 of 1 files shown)


config/grants.yml
```diff
@@ -196,3 +196,8 @@
 # Allow the taskcluster team to handle the denylist
 - grant: notify:manage-denylist
   to: github-team:taskcluster/core
+
+# Make all endpoints public by default
+- grant: assume:anonymous
+  to: *
```
djmitche, line 202: Granting `assume:anonymous` to everything is what the RFC#165 PR does :)

What you want to do here is `grant:` a bunch of scopes `to:` the `anonymous` role:

```yaml
- grant:
  - scope
  - scope
  to: anonymous
```

djmitche (changes requested):

config/grants.yml
```diff
@@ -196,3 +196,47 @@
 # Allow the taskcluster team to handle the denylist
 - grant: notify:manage-denylist
   to: github-team:taskcluster/core
+
+# Make all endpoints public by default
```
djmitche, line 200:
```suggestion
# Make read-only endpoints public by default
```
instead, maybe?

djmitche (approved):

PR #355, opened 2020-10-27
Default to not having video-loopback in community d-w workers

imbstack:


djmitche (approved):
Just need to fix the lint error.

djmitche:
(partially applied along with #356)

PR #356, opened 2020-10-30
Add grizzly-reduce hooks

jschwartzentruber:
Grizzly is a framework used to run all of our DOM fuzzers. This adds some hooks that will be used for reducing Grizzly testcases to their minimum reproducible size.

`grizzly-reduce-monitor` will eventually run on a schedule, but for now having the hooks in place will aid with testing the underlying images.


djmitche (approved):
Thanks for the explanation :)

djmitche:
applied

PR #357, opened 2020-10-31
New images oct 30 2020

imbstack:
If there's a better way to go about disabling devices in aws only, let me know! This feels like a hack.


djmitche (approved):
I think this makes sense.  It might do to include a comment indicating that these are enabled by default (so, on gcp).

PR #358, opened 2020-11-04
Add a proj-misc/tutorial pool

djmitche:
This pool has a maxCapacity of 2, and is available to anyone with a GitHub login.  That should minimize the scope of any abuse, and at least give us a user to which to trace the abuse.

### DIFF
--- a/config/projects/misc.yml
+++ b/config/projects/misc.yml
@@ -15,6 +15,14 @@ misc:
       cloud: gcp
       minCapacity: 0
       maxCapacity: 20
+      # basically everyone has access to this worker (see grants below)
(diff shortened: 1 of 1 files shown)


imbstack (approved):
it makes me a bit uncomfortable and we may regret this next time we have a security issue but it does seem nice to have

djmitche:
Unless that security issue allows compromise of the entire GCP project, I think we're OK.  Maybe someone mines a little bitcoin on those two n1-standard's until we get grumpy about it -- not too bad.  And we'd know who did it (at least a github userid).

djmitche:
(applied)

PR #361, opened 2020-11-12
Change fuzzing CI scope to allow individual pypi tokens.

jschwartzentruber:


djmitche (approved):
This change will drop scopes for the "plain" secret.  Is that OK?

jschwartzentruber:
Yes, that's intentional. It's only in use in one repository and I've already added a more specific secret for it.

djmitche:
OK, applied!

Issue #367, opened 2020-12-09
Requesting access to Community instance for testing


MasterWayZ:
Hi,

Forwarded here from the Taskcluster channel on Matrix.

I'm really interested in Taskcluster and would like to explore it and learn it, testing things like workers and jobs. Since there are no docs yet on how to set up a Taskcluster instance yourself, I was told I could possible get access to the community instance for it, which is why I'm making this issue.

If you need any more information, please let me know.

djmitche:
At a very basic level, you can already sign in and create a task at https://community-tc.services.mozilla.com/tasks/create -- all github users have access to the "proj-misc/tutorial" worker pool (which has pretty limited capacity).

This repository configures permissions for "projects" which are basically protected namespaces within the deployment.  You can see an example of a project for [RTEMS](https://github.com/mozilla/community-tc-config/blob/main/config/projects/rtems.yml) which was at one point interested in trying out Taskcluster (/cc @verm).  The result was that Amar got a bunch of scopes when they logged in, including scopes to create workers.  A more sophisticated project is [servo](https://github.com/mozilla/community-tc-config/blob/main/config/projects/servo.yml) which defines some worker pools and some permissions for specific github repos.

From what I understand, you're interested in experimenting with workers.  Deploying workers is necessarily a bit complicated, but maybe the best way to get you the permissions you need to do that is to set up a new project for you similar to RTEMS.  That will give you, among other things, `worker-manager:manage-worker-pool:proj-yourproject/*`, allowing you to create worker pools via the API or the UI.  What should that project be called?

MasterWayZ:
Sounds good! I've looked into workers and I think I will need some help with it.
My general idea for this was to have a worker that I can bring myself in a VM, to run commands in a docker container. Like pulling the repo, building files and then deploying them, or even that plus having it as an artifact if that is a thing. This is just an idea I had to learn and experiment with workers and Taskcluster. One thing is that the repo is on a GitLab server, and not on GitHub. But if it's needed I can duplicate it over to GitHub. (Not sure if there's a way to get it to pull from a GitLab server)

The project name can be mwz or masterwayz, depending on what it easier to use.

djmitche:
That sounds quite possible!  We can also create a worker pool for you that will automatically create a docker-worker instance in GCP or AWS.  But it sounds like you want to do the deployment yourself :)

[...]

PR #368, opened 2020-12-09
Add an mwz project to support experimentation

djmitche:
Refs #367.


djmitche:
I added a pool with provisionerId `proj-mwz` and workerType `testing`.  That has a max capacity of 1, so only one task at a time.  It will be started as needed.

imbstack (approved):

djmitche:
Applied!  If you login to https://community-tc.services.mozilla.com/profile you should see a bunch of scopes to allow you to edit resources with a `mwz` prefix.

MasterWayZ:
I do yes, thank you!

PR #373, opened 2020-12-17
add blanket proj-misc access to the mozilla-services org

djmitche:
This org just got notices to move away from travis-ci, so let's make the transition to TC a little easier.


imbstack (approved):
seems reasonable to me

PR #374, opened 2020-12-19
Grant assume:project:bugbug/deploy to all bugbug hooks

marco-c:


djmitche (approved):
I'm at my personal laptop and can't apply this right now.  Is it OK for this to wait until Monday?

djmitche:
Applied!  I see now that the grant had already been applied, so it seems I didn't slow you down :)

marco-c:
Yes, thanks!

PR #375, opened 2020-12-21
Add more repositories to the relman configuration

marco-c:
All repos that were using Travis CI and are going to use Taskcluster instead.


djmitche (approved):

djmitche:
applied!

PR #382, opened 2021-01-12
Add more unapplied changes for fuzzing.

jschwartzentruber:
This should be the rest of the live changes. The role changes are intended to simplify things a bit by making some grants common.


djmitche (approved):

djmitche:
applied with only the role diff you indicated

djmitche:
thanks!!

PR #384, opened 2021-01-12
Fix routes in fuzzing hooks.

jschwartzentruber:


jschwartzentruber:
Should tc-admin raise an error here?

djmitche:
Unfortunately not -- the incorrect `routes` property was part of the task template, which is just an opaque JSON-e "program" until the hook is triggered.

djmitche:
I got that completely backward.  Yes, it should.  I'll file an issue.

djmitche (approved):

PR #385, opened 2021-01-25
add queue:create-task:project:none to Role=anonymous

djmitche:
For taskcluster/taskcluster#4262.


petemoore (approved):
nit: it might be nice to add this a couple of lines higher up instead so that the scopes remain in sorted order

djmitche:
I'll fix that and merge

PR #392, opened 2021-02-03
Add scopes for permission elevation in servo tasks.

jdm:


djmitche:
@petemoore can you check this?

petemoore (approved):
This is fine so long as tasks are reasonably trusted. Running as administrator will mean they could potentially access taskcluster credentials of the worker (e.g. generic-worker config file).

@jdm Is this acceptable?

petemoore:
You've probably realised this already, but you'll also need to include the following in the task definitions:

```
...
payload:
  osGroups:
    - Administrators
  features:
    runAsAdministrator: true
```

If I remember correctly, if you have UAC disabled on the workers (i.e. you disabled the feature in Windows), the `runAsAdministrator: true` should not be set. But usually UAC is enabled by default, so this is unlikely.

jdm:
Yes, these scopes will only be used by trusted tasks, and I've added those payload features as well. Thanks for checking!

djmitche:
I'll merge on Pete's OK then :)

djmitche:
```
rubin ~/p/community-tc-config [main] $ tc-admin apply --without-secrets --grep servo
Updating Role=project-admin:servo
```

PR #393, opened 2021-02-17
Add missing scopes for fuzzing.

jschwartzentruber:


djmitche (approved):
Y'all have the best names!

djmitche:
(it looks like this is already applied)

PR #398, opened 2021-03-15
Update secrets for gr-css

pyoor:
Several secrets were renamed to match our existing naming convention and grant wild cards.  This PR updates those secrets for gr-css.


djmitche (approved):

djmitche:
(looks like this was already applied, so I didn't do anything)

PR #402, opened 2021-04-19
[fuzzing] Increase max CI pool to 20 and grant cancel-task permission for project-fuzzing to fuzzing admins.

jschwartzentruber:
The reason for the CI increase is that I'm using that queue now for all fuzzing decisions instead of running decisions in the fuzzing work queue.

### DIFF
--- a/config/projects/fuzzing.yml
+++ b/config/projects/fuzzing.yml
@@ -25,7 +25,7 @@ fuzzing:
       imageset: docker-worker
       cloud: gcp
       minCapacity: 0
-      maxCapacity: 10
+      maxCapacity: 20
       workerConfig:
         dockerConfig:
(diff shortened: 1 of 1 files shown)


config/projects/fuzzing.yml
```diff
@@ -408,6 +408,7 @@ fuzzing:
         - repo:github.com/MozillaSecurity/grizzly-reduce-tc:*
     - grant:
         - generic-worker:allow-rdp:proj-fuzzing/*
+        - queue:cancel-task:fuzzing/*
```
djmitche, line 411: 🎈

djmitche (approved):

djmitche:
(applied)

PR #405, opened 2021-05-03
Add missing scopes for js-distiller.

jschwartzentruber:


djmitche (approved):

djmitche:
(it looks like this was already applied)

PR #406, opened 2021-05-05
Update tc-admin to 3.0.0

djmitche:


setup.py
```diff
@@ -9,7 +9,7 @@
     url="https://github.com/djmitche/community-tc-config",
     packages=find_packages("."),
     install_requires=[
-        "tc-admin~=2.5.1",
+        "tc-admin>=3.0.0",
```
imbstack, line 12: It's been a while since I knew a lot about this stuff but is there a reason we don't want to keep `~=`?

imbstack (approved):

setup.py
djmitche: Good point; I'm not sure why I did that

PR #407, opened 2021-05-07
Fuzzing config changes

jschwartzentruber:
- allow Windows fuzzing tasks to run as Administrator
- remove scopes for archived repos
- add scopes for FuzzManager to build/publish Docker images

### DIFF
--- a/config/projects/fuzzing.yml
+++ b/config/projects/fuzzing.yml
@@ -441,6 +441,14 @@ fuzzing:
       to:
         - repo:github.com/MozillaSecurity/orion:*
     - grant:
+        - docker-worker:capability:privileged
+      to:
+        - repo:github.com/MozillaSecurity/FuzzManager:*
+    - grant:
+        - secrets:get:project/fuzzing/docker-hub
+      to:
+        - repo:github.com/MozillaSecurity/FuzzManager:branch:master
+    - grant:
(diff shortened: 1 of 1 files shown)


petemoore (commented):
Hey @jschwartzentruber! Have you been able to test these changes? I had a look at https://community-tc.services.mozilla.com/worker-manager/proj-fuzzing%2Fci-windows but it looks like they are still using generic-worker-win2012r2. I'm not sure if the win2016 worker type images are well maintained or even working - so just wanted to check if you've been able to stage/test the changes, ideally before we merge/deploy.

jschwartzentruber:
Yes, I launched an interactive task in a pool proj-fuzzing/truber-win-test, which looks like it has disappeared now. RDP and fuzzing worked fine. I've since moved the fuzzing pool over (https://community-tc.services.mozilla.com/worker-manager/proj-fuzzing%2Fwindows-pool55) and haven't seen any problems yet related to the image. The only reason I moved from win2012r2 to win2016 is that it's more familiar and easier to find documentation.

The problem we're facing (which is the reason for the `Administrator` scopes) is that we need to add a Defender exception for fuzzing temp folders, if Defender is enabled in the image. Otherwise we end up with poor performance (since Defender scans everything we create) and file locking issues (trying to rapidly create/delete temp files while Defender is still scanning).

PR #408, opened 2021-05-11
Fix fuzzing Administrators scope. 

jschwartzentruber:
Wildcards are only valid against suffixes.


imbstack (approved):
I'll apply this shortly

djmitche:
applied

PR #410, opened 2021-06-14
Dustin offboarding

petemoore:
:cry:

### DIFF
--- a/config/projects/mwz.yml
+++ b/config/projects/mwz.yml
@@ -1,7 +1,6 @@
 # https://github.com/MasterWayZ is experimenting with TC workers and jobs
 mwz:
   adminRoles:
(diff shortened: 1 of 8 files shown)


config/projects/mwz.yml
```diff
@@ -1,7 +1,6 @@
 # https://github.com/MasterWayZ is experimenting with TC workers and jobs
 mwz:
   adminRoles:
-    - login-identity:github/28673|djmitche
```
jwhitlock, line 4: We talked about this in the team meeting. These experimental projects don't need a Mozilla employee as an admin at the present, and can be added in the future if needed.

jwhitlock (approved):
Thanks @petemoore!

PR #415, opened 2021-09-14
[fuzzing] Add hook for bugmon forced confirmation

pyoor:
Adds a hook for running the bugmon image weekly with the `--force-confirm` flag.


jwhitlock (approved):
Thanks @pyoor, looks good to me, and that it is already applied in community-tc (except for the description)

PR #420, opened 2021-09-16
Added README.md section on how community-tc admins should apply changes

petemoore:
Make it clear that root credentials are typically needed to apply changes.

Note, we could probably create a more restricted role that the taskcluster team gets, that allows creation of arbitrary entities that tc-admin controls, but doesn't have scopes to e.g. create tasks.

Or we could have a separate project for community-tc admins, that is different to the taskcluster project, and grant scopes to the community-tc-admins to be able to create all the entities.

But for now I think this is a reasonable solution, since people that run tc-admin should typically have access to the root access token anyway.


jwhitlock (approved):
Thanks for clarifying this @petemoore!

PR #424, opened 2021-09-22
Changes needed for firefox-ci docker-worker images

petemoore:
This wasn't needed in community, but was needed in firefox-ci - so keeping track of changes here, so we have something more portable.


jwhitlock (approved):
Thanks @petemoore!

PR #427, opened 2021-09-27
Bootstrap cert from file

petemoore:


jwhitlock (approved):
👍

petemoore:
After this landed, I realised it won't work, since only the bootstrap.sh script is on the host, and not other files from this repo. Therefore I'm going to need to revert this change.
Note the cert patching is only a temporary measure, soon we hope to replace docker-worker, and then we won't need the cert at all (generic-worker doesn't require it for live logging). Therefore there probably isn't a lot of value in adding a check for the cert expiry, as by the time the current cert expires, we should have moved off docker-worker, or be very close to the finish line for doing so.

PR #428, opened 2021-09-28
Default AWS instance type m3.2xlarge -> m4.2xlarge, and GCP machine types n1 -> n2

petemoore:
I'm not an expert here, but I think there is a shortage of m3.2xlarge instances, which is why we have recently been receiving Worker Manager emails saying that it can't launch (mostly) [proj-taskcluster/gw-ci-ubuntu-18-04](https://community-tc.services.mozilla.com/worker-manager/proj-taskcluster%2Fgw-ci-ubuntu-18-04) instances.

I looked at the _General Purpose_ instance types [here](https://aws.amazon.com/ec2/instance-types/) and M3 wasn't even listed in the table. I suspect M4 is the closest choice.

I'm not sure how this will affect spot pricing, or even how to measure any cost implications this has. Therefore I'm keen to have someone else chip in with their thoughts, rather than just pushing this change through.

I also saw there is already [code](https://github.com/mozilla/community-tc-config/blob/0822ba47d08a1aab7eca612304b612f88f249798/generate/workers.py#L365-L371) that checks if the instance type is m3.2xlarge and doesn't launch in a list of us-east availability zones, so I guess we hit this with us-east before. The newer failures seem to be mostly us-west availability zones.

Furthermore, @glandium has reported that [proj-git-cinnabar/win2012r2](https://community-tc.services.mozilla.com/worker-manager/proj-git-cinnabar%2Fwin2012r2) instances are not getting launched, and I see that this worker type also uses the default m3.2xlarge instance size - so this may be the cause of his issues too.

PR #431, opened 2021-09-30
Default ec2 instance type m4.2xlarge => m5.2xlarge

petemoore:
m4.2xlarge is pretty old, and is often unavailable, therefore changing default instance type to m5.2xlarge. See also https://github.com/mozilla/community-tc-config/pull/428#issuecomment-929066262.

Originally I [intended to update](https://github.com/mozilla/community-tc-config/pull/428#issuecomment-930130397) the hardcoded [EC2 availability zones](https://github.com/mozilla/community-tc-config/blob/266a6c02d7d3e6bf86bd192a986355c18266fb23/generate/workers.py#L365-L371)/[GCP zones](https://github.com/mozilla/community-tc-config/blob/266a6c02d7d3e6bf86bd192a986355c18266fb23/generate/workers.py#L269-L278) in this PR. However, I've since seen that I can use [`aws ec2 describe-instance-types`](https://docs.aws.amazon.com/ko_kr/cli/latest/reference/ec2/describe-instance-types.html) and [`gcloud compute machine-types list`](https://cloud.google.com/sdk/gcloud/reference/compute/machine-types/list) to determine if an instance type / machine type is available in a given EC2 availability zone / GCP zone, which is a bit more work, but should require less maintenance. Therefore I'll put that in its own PR.

### DIFF
--- a/generate/workers.py
+++ b/generate/workers.py
@@ -319,7 +319,7 @@ def aws(
     *,
     image_set=None,
     regions=None,
-    instanceTypes={"m4.2xlarge": 1},
+    instanceTypes={"m5.2xlarge": 1},
     securityGroup="no-inbound",
     minCapacity=0,
     maxCapacity=None,


jwhitlock (approved):
👍

Issue #432, opened 2021-09-30
Remove hardcoded blacklists of instance/machine type to (availability) zones


petemoore:
Originally from https://github.com/mozilla/community-tc-config/pull/428#issuecomment-930130397.

There are hardcoded blacklists of [EC2 availability zones](https://github.com/mozilla/community-tc-config/blob/266a6c02d7d3e6bf86bd192a986355c18266fb23/generate/workers.py#L365-L371)/[GCP zones](https://github.com/mozilla/community-tc-config/blob/266a6c02d7d3e6bf86bd192a986355c18266fb23/generate/workers.py#L269-L278) for a given instance type / machine type, that can be replaced by calls to [`aws ec2 describe-instance-type-offerings`](https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-instance-type-offerings.html) and [`gcloud compute machine-types list`](https://cloud.google.com/sdk/gcloud/reference/compute/machine-types/list) to determine if an instance type / machine type is available in a given EC2 availability zone / GCP zone. This is desirable, since this hardcoded list is constantly subject to change, but the API calls that would replace them should not need to change over time.

petemoore:
One consideration here, is that currently tc-admin does not require AWS/GCP credentials and does not call any AWS/GCP apis. Ideally we'd find a way to perform these lookups without needing protected credentials. I'm not sure if AWS/GCP have public endpoints that don't require credentials, or if we would be forced to create credentials/service account that only had the permissions to perform the instance type / machine type lookup calls described above.

petemoore:
@jwhitlock What are your thoughts?

petemoore:
I guess we could have an option like `--update-instance-types-per-zone` which updates JSON files in the repo with the mappings, which is not run by default. Then when things change, we can run it with the option, it will update the repo files, which are then of course persisted, so then you have the choice of running it or not. If you don't specify the argument, the old mappings are not updated. Alternatively we could invert the logic, so you have to specify `--do-not-update-instance-types-per-zone` to not update them - I think I would prefer that so you have to explicitly opt in to use cached values. If you don't have valid AWS/GCP creds then you will get an error message to say to either specify them, or provide this option not to update the mappings.

petemoore:
Possible solution:

[...]

PR #435, opened 2021-10-04
Issue #432 - work in progress

petemoore:
Initially using the AWS CLI to generate the mappings, we could consider using the python client instead.

We might also want to filter just on the availability zones that are referenced and/or the instance types that are referenced, in order to have a smaller footprint - although having everything can be useful for tracking changes over time, and also maybe keeps things more simple.


jwhitlock:
I've added code to `generate/workers.py` to use these JSON files to dynamically filter non-available worker types by zone. I've also added an assertion that at least one launch config is generated, to avoid pools with no launch configurations. I could add a second check that each region (like "us-west-1") also has at least one launch config.

petemoore:
> I've added code to `generate/workers.py` to use these JSON files to dynamically filter non-available worker types by zone. I've also added an assertion that at least one launch config is generated, to avoid pools with no launch configurations. I could add a second check that each region (like "us-west-1") also has at least one launch config.

This is awesome, many thanks John! I noticed that `m5.2xlarge` is only supported in two availability zones of the regions worker-manager uses (`us-west-1`, `us-west-2`, `us-east-1`), with both of them being in `us-west-1` (`us-west-1a` and `us-west-1b`). However there are several other regions that do support this instance type, that Worker Manager doesn't use:

```
ap-northeast-1
ap-northeast-2
ap-northeast-3
ap-south-1
ap-southeast-1
ap-southeast-2
ca-central-1
eu-central-1
eu-north-1
eu-west-1
eu-west-2
eu-west-3
sa-east-1
us-east-2
```

I'm not sure why Worker Manager only uses `us-east-1`, `us-west-1` and `us-west-2` but perhaps we should review that decision.

petemoore:
@jwhitlock I've renamed things a little, cleaned up a typo, created new commits, and added the google cloud machine types to the `/config` directory. The format is a little different from the amazon instance types; there is a single file rather than one per zone. The data is structured as a list of (machine type, zone) pairs (which was easier to create than e.g. a mapping from zone to machine types). Do you think this is ok? We still need to update [`machine_in_zone(machineType, zone)`](https://github.com/mozilla/community-tc-config/blob/266a6c02d7d3e6bf86bd192a986355c18266fb23/generate/workers.py#L269-L278) to use this new data source.

petemoore:
Thanks for your commits John! I don't think I can approve the PR since I created it, but I'm happy to confirm here in writing that I r+ your commits! Hopefully GitHub still allows to review the PR even though you have contributed some of the commits. If not we'll have to rope somebody else in.

jwhitlock (approved):

PR #439, opened 2021-10-11
Explicitly set background console colour to contrast foreground colour

petemoore:
When using a terminal with a white background colour I could not clearly read text; by explicitly setting background colour the console output should be clear in all terminals now.


jwhitlock (approved):
🤯

PR #440, opened 2021-10-11
Ensure yq version 3 is used by imageset.sh

petemoore:
This requirement was [already documented](https://github.com/mozilla/community-tc-config/blob/3ae9b2c05e540a9a2f35336d1058da2e7c8fe8b2/imagesets/README.md?plain=1#L120), but not included in checks.

### DIFF
--- a/imagesets/README.md
+++ b/imagesets/README.md
@@ -117,7 +117,7 @@ All of the following tools must be available in the `PATH`:
   * `tr`
   * `which`
   * `xargs`
-  * `yq` **version 3** (version 4 not supported)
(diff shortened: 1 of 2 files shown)


petemoore:
> Seems good to me. Do we know what's blocking us from supporting version 4 of `yq`?

Indeed that is useful context we should probably include. The command syntax changed between version 3 and 4, and we are still on the v3 syntax. See [here](https://mikefarah.gitbook.io/yq/upgrading-from-v3).

An alternative solution is that we upgrade to v4, but either way we would need to check the version matched.

PR #441, opened 2021-10-12
Add region us-east-2 for building AWS EC2 worker AMIs

### LINKED ISSUE #438: Generate AWS region us-east-2 launch configurations
The AWS EC2 worker pools are currently limited to regions us-east-1 / us-west-1 / us-west-2, which often results in Worker Manager failing to launch instances due to limited supply.

By adding the us-east-2 availability zones to the worker pool definitions, supply should be improved, and we should have better coverage when supply is low in the current regions. In particular, `m5.2xlarge`, which is our default instance type, is only supported in availability zones us-west-1a and us-west-1b with the currently configured regions, but is supported in many us-east-2 availability zones, so would provide a lot more coverage if us-east-2 was supported.

Note, before we add us-east-2 availability zones in the worker pool definitions, we will first need to add support for us-east-2 to the community cluster Worker Manager, which may involve ec2 configuration (e.g. creation of security groups, vpn configuration, IAM configuration, ...) and possibly additional Worker Manager configuration(?)

petemoore:
Adding us-east-2 to aws regions to build AMIs in.

Fixes #438.

### DIFF
--- a/config/imagesets.yml
+++ b/config/imagesets.yml
@@ -79,9 +79,10 @@ generic-worker-win2012r2-staging:
   workerImplementation: generic-worker
   aws:
     amis:
-      us-east-1: ami-082c2d64e8209bb0f
-      us-west-1: ami-055c72aff254354fa
-      us-west-2: ami-0820db6bf4ff3c943
+      us-east-1: ami-0645f31a90010e574
+      us-west-1: ami-01ecc4a44b76f7863
+      us-west-2: ami-0f984fd4c43810fe0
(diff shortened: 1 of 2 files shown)


petemoore:
Task claimed successfully! Payload was invalid, but that was intentional, the test was only that a worker would be spawned and that it could claim and resolve a task, which it did. Log [here](https://community-tc.services.mozilla.com/tasks/QotHkQveRqO1RJunXc4p0w/runs/0/logs/public/logs/live_backing.log).

PR #443, opened 2021-10-12
Run imageset.sh from arbitrary dir

petemoore:
There was a bug that you could only run `imageset.sh` from the current directory (i.e. `./imageset.sh .....`). It didn't work if you ran it from e.g. a different working directory (e.g. if you ran `imagesets/imageset.sh` from the root directory of the repo).

This has been fixed. Note this may create a merge conflict with #441 since they both change the same line, but we can fix that when we merge.


jwhitlock (approved):
Thanks @petemoore, I think you got all of them!

PR #450, opened 2021-11-12
add support for the json-e/json-e repository

djmitche:
See https://github.com/json-e/json-e/issues/430

This leaves JSON-e in the "taskcluster" project.  I think that will be easiest for the moment.

### DIFF
--- a/config/projects/taskcluster.yml
+++ b/config/projects/taskcluster.yml
@@ -3,6 +3,7 @@ taskcluster:
     - github-team:taskcluster/core
   repos:
     - github.com/taskcluster/*
+    - github.com/json-e/json-e:*
     - github.com/mozilla/community-tc-config:*
     - github.com/mozilla/hawk:*
   externallyManaged:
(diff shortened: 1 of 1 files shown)


jwhitlock (approved):
Thanks @djmitche!

It has this scope with a notification:

```
scopes:
      - queue:route:notify.email.taskcluster-notifications@mozilla.com.*
      - secrets:get:project/taskcluster/json-e-deploy
```
I'm not sure if this means "send this email", or if it just allows sending the email. There may be a way to email the json-e maintainers instead

jwhitlock:
Applied to https://community-tc.services.mozilla.com/

djmitche:
Thanks!  We'd have to make a mailing list for the maintainers, which is probably not worth the trouble.

jwhitlock:
I thought the Github teams had an email feature, but it is a team discussion feature that is on the website and doesn't have a mailbox. I agree, probably not worth the trouble.

PR #453, opened 2021-11-22
Fix syntax to use indexed docker image in mozci hook

La0:
I tested that syntax on a [sample task](https://community-tc.services.mozilla.com/tasks/B1x_y123SPCGRFtYpJvCrQ/runs/0/logs/live/public/logs/live.log)


petemoore (approved):

petemoore:
Updated hook `project-mozci/decision-task-testing`

PR #455, opened 2021-12-02
Grant access to docker-worker cache for mozci decision tasks

La0:
Needed for https://github.com/mozilla/mozci/issues/564

I'm setting up a shared cache for the mozci classification tasks. Pickled payloads using [cachy](https://github.com/sdispater/cachy/) will be stored on that cache : we do not expect a lot of disk storage needed (we do not clone MC ...)

I also need access to a secret in order to configure cache settings for the different tasks.

### DIFF
--- a/config/projects/mozci.yml
+++ b/config/projects/mozci.yml
@@ -57,6 +57,9 @@ mozci:
         # Children tasks are indexed
         - queue:route:index.project.mozci.*
 
+        # Children tasks need to share a common cache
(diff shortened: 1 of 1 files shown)


petemoore (approved):

petemoore:
Deployed, and role [hook-id:project-mozci/decision-task-testing](https://community-tc.services.mozilla.com/auth/roles/hook-id%3Aproject-mozci%2Fdecision-task-testing) was updated. :-)

PR #462, opened 2022-02-03
Use correct mozci date parameter in date hook

La0:
Sorry, again a fix on the mozci hook, we had a communication issue on that parameter :cry: 

Needed to fix the current error on the hook https://community-tc.services.mozilla.com/tasks/QmrFXEhySYWroLSi6C9cvg/runs/0/logs/public/logs/live.log

### DIFF
--- a/config/projects/mozci.yml
+++ b/config/projects/mozci.yml
@@ -122,7 +122,7 @@ mozci:
           command:
             - push
             - classify-eval
-            - "--from=1 days ago"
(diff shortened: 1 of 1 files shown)


EvaBardou:
@La0 Fixed syntax tested on my local instance and it works perfectly :+1:

petemoore (approved):

petemoore:
Landed...

```
Updating Hook=project-mozci/monitoring-testing
```

PR #463, opened 2022-02-07
[fuzzing] Create a pool for Windows testcase reduction.

jschwartzentruber:


jwhitlock:
Applied with:

```
> tc-admin apply --without-secrets --ids-only --grep "fuzzing|Security"
Updating Hook=project-fuzzing/grizzly-reduce-monitor
Creating Role=worker-pool:proj-fuzzing/grizzly-reduce-worker-windows
Updating Role=repo:github.com/MozillaSecurity/*
Creating WorkerPool=proj-fuzzing/grizzly-reduce-worker-windows
```

PR #467, opened 2022-02-22
Fix mozci scopes for Matrix notifs + Add 2 flags to monitoring hooks

EvaBardou:
Related to https://github.com/mozilla/mozci/issues/647 and https://github.com/mozilla/mozci/issues/608

### DIFF
--- a/config/projects/mozci.yml
+++ b/config/projects/mozci.yml
@@ -64,8 +64,9 @@ mozci:
         - notify:email:*
 
         # Children tasks are able to send notifications
-        # to code sherriffs through Matrix
-        - notify:matrix-room:!sheriff-notifications:mozilla.org
+        # to code sheriffs through Matrix
+        - notify:matrix-room:#sheriff-notifications:mozilla.org
+        - notify:matrix-room:!vNAdpBnFtfGfispLtR:mozilla.org
       to:
         - hook-id:project-mozci/decision-task-testing
         - hook-id:project-mozci/decision-task-production
@@ -153,6 +154,8 @@ mozci:
             - classify-eval
             - "--from-date=1 days ago"
             - --send-email
+            - --recalculate
+            - --detailed-classifications
             - --environment=testing
           maxRunTime: 1800
         scopes:
@@ -213,6 +216,8 @@ mozci:
             - classify-eval
             - "--from-date=1 days ago"
             - --send-email
+            - --recalculate
+            - --detailed-classifications
             - --environment=production
           maxRunTime: 1800
         scopes:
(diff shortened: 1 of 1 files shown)


marco-c (approved):

config/projects/mozci.yml
```diff
@@ -64,8 +64,9 @@ mozci:
         - notify:email:*
 
         # Children tasks are able to send notifications
-        # to code sherriffs through Matrix
-        - notify:matrix-room:!sheriff-notifications:mozilla.org
+        # to code sheriffs through Matrix
+        - notify:matrix-room:#sheriff-notifications:mozilla.org
+        - notify:matrix-room:!vNAdpBnFtfGfispLtR:mozilla.org
```
petemoore, line 69: This middle line (68) is not needed - the internal room ID on line 69 (`!vNAdpBnFtfGfispLtR:mozilla.org`) is the correct way to refer to the `#sheriff-notifications` room.

However, I suspect the scope needed is probably `queue:route:notify.matrix-room.!vNAdpBnFtfGfispLtR:mozilla.org` unless the hook calls the notify API at runtime from the task. If it notifies the channel by including the task route `notify.matrix-room.!vNAdpBnFtfGfispLtR:mozilla.org` in the task definition, then `queue:route:notify.matrix-room.!vNAdpBnFtfGfispLtR:mozilla.org` is instead the required scope.

It would be nice to change the comment from `to code sheriffs through Matrix` to something like `to #sheriff-notifications room on Matrix` (just to be a bit more explicit).

Many thanks!

petemoore (changes requested):
It is close, but just needs a little tweaking. Thanks!

config/projects/mozci.yml
La0: We are using the notify API at runtime from the task. Here is an example of [currently failing task](https://community-tc.services.mozilla.com/tasks/ZAwPzeUNSkCbZg09eu8mfg/runs/0/logs/public/logs/live.log) which mention the internal Matrix room ID.

I think @EvaBardou added both values to match what is done on the [servo](https://github.com/mozilla/community-tc-config/blob/main/config/projects/servo.yml#L32) project

EvaBardou: Thanks @La0 for your reply. :smile: 
@petemoore I updated the comment and deleted the unnecessary/erroneous scope from line 68.

petemoore (approved):
Looks great, many thanks!

petemoore:
Changes applied:

```
Updating Hook=project-mozci/monitoring-testing
Updating Role=hook-id:project-mozci/decision-task-production
Updating Hook=project-mozci/monitoring-production
Updating Role=hook-id:project-mozci/decision-task-testing
```

PR #471, opened 2022-03-21
Fix imageset generation on macOS

petemoore:
This is a bit of a cleanup for the issues we discovered today. It does the following:

1) `signin-aws` is now not a system dependency, but is included in the repository. It has been renamed to `signin-aws.sh` to make it clearer that it is just a shell script. Also the name doesn't overload the function name that can be used to source it from the shell, to make it a little more obvious that the function and the executable script are different things.

The idea is that if you source its output, it sets env vars for you for 6 hours, so you don't need to run it every time you build an image set, e.g. you could source its output once to get an AWS session, and then create all image sets, without needing to reauthenticate.

2) The unique ID generation process has changed from using `tr` to using `base64` instead. It achieves a similar goal, in a slightly more portable way - essentially generated a random string of 20 chars including [a-z0-9] chars. Since some filesystems are case insensitive, I went with lowercase a-z only, to keep things simple.

3) I rebuild the docker image and bumped the version. Also explicitly download v3.4.1 of yq instead of latest version in Dockerfile.

4) I removed dependency on grep, since sed is already a dependency, used that instead.

5) Made the detection of exceptions more strict and consistent with

```
set -eu
set -o pipefail
```

I also tried adding `export SHELLOPTS` but this broke the `pass` commands, and then on reflection, I decided it was better not to have it. The alternative would be to add it, and then unexport SHELLOPTS (`export -n SHELLOPTS`) prior to any `pass` commands (and `export` it again after each `pass` command) - but that felt a bit overkill. Another option would be to fix the environment so that `SHELLOPTS` could be exported (e.g. `PASSWORD_STORE_GPG_OPTS= pass ...`) but then it becomes very sensitive to `pass` changes, so on reflection I decided better not to try to be too clever.

It is possible that this might introduce a failure where we didn't have one before, but since we are going to be testing rollout of AWS and Google images, hopefully we'll catch anything that this might have broken (e.g. due to an environment variable being evaluated that hasn't been initialised). Also I'm currently testing a deployment of `generic-worker-ubuntu-18-04-staging` into AWS to see if all is ok.

6) More consistent syntax (e.g. `${var}` rather than `$var`) when I saw inconsistencies.

### DIFF
--- a/config/imagesets.yml
+++ b/config/imagesets.yml
@@ -135,9 +135,10 @@ generic-worker-ubuntu-18-04-staging:
   workerImplementation: generic-worker
   aws:
     amis:
-      us-east-1: ami-0bfbcea1e13dfdb8a
-      us-west-1: ami-0b9d053fd6a809c8c
-      us-west-2: ami-0f402ebee06e1d5c9
+      us-east-1: ami-06bf86c730ce8a38d
+      us-west-1: ami-097c7e3654aad8eb6
+      us-west-2: ami-034872d3c9919b020
+      us-east-2: ami-0944cf237a1b66d99
   workerConfig:
     genericWorker:
       config:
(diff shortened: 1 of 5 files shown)


lotas (approved):
:+1:

PR #477, opened 2022-03-23
Some fix-ups to imageset.sh

petemoore:
The `imageset.sh` script builds machine images (expensive, time-consuming) and then updates password store with credentials to connect to new machine images. However, if there is a problem with the password store / gpg setup of the user running `imageset.sh`, they will only discover this after the new machine images have been built.

I added a pre-check that writing secrets to the password store works, before beginning the long process of generating machine images, to reduce chance of failure once the image building process has started.

At the same time, I modified the log routine to clear to end-of-line on line return, so that the background colour extends to the full terminal width on line returns (aesthetic improvement).

I also moved all checks to before the 'Starting!' log line and added a check that the specified imageset directory exists.

Note:

One thing I don't love about the new password store test, is that if for some reason the test secret insertion is successful, but the test secret removal isn't successful, there is a risk that we get a secret `test` added to the password store.

However, I think the risk outweighs the benefit. I'm not sure of the best way to guard against this. Rerunning the script would fetch a new clone of the repo, so the only way this could be a problem is if the removal of the secret silently failed, and then the script continued to build images and push changes to the git repo. This seems like a very small risk, as the secret removal silently failing would indicate a bug in pass or git (`pass git reset --hard HEAD~1`) since the pass command should have a non-zero exit command if the command fails. Also since it doesn't interact with the network, there is low chance it should fail. It is also a dedicated checkout of the password store for the current process, rather than a system-wide password store checkout, so there shouldn't be any race conditions which could cause the wrong commit to be removed etc.

### DIFF
--- a/imagesets/imageset.sh
+++ b/imagesets/imageset.sh
@@ -5,9 +5,9 @@ set -o pipefail
 
 function log {
   if [ -n "${BACKGROUND_COLOUR-}" ] && [ -n "${FOREGROUND_COLOUR-}" ] && [ -n "${CLOUD-}" ] && [ -n "${IMAGE_SET-}" ] && [ -n "${REGION-}" ]; then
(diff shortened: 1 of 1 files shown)


petemoore:
Whoops - landed this accidentally, then reverted it. Creating a new PR for this.

PR #484, opened 2022-03-25
Generic worker on Windows Server 2022

petemoore:
Let's see what works when deploying generic-worker multiuser engine to Windows Server 2022 Base in AWS/EC2.

If all is well, we should look at migrating all our workloads from Windows Server 2012 to Windows Server 2022.


matt-boris (approved):
Awesome! 🎉

PR #486, opened 2022-04-01
Fetch root credentials from pass when running tc-admin

petemoore:
Note this isn't tested by me, since my python environment is messed up - can you check it works for you Yarik?

Thanks!

### DIFF
--- a/imagesets/README.md
+++ b/imagesets/README.md
@@ -147,14 +147,19 @@ complete:
    pip3 install -e .
    which tc-admin
    pass git pull
-   TASKCLUSTER_ROOT_URL=https://community-tc.services.mozilla.com tc-admin diff || true
+   TASKCLUSTER_ROOT_URL=https://community-tc.services.mozilla.com
+   TASKCLUSTER_CLIENT_ID=static/taskcluster/root
+   TASKCLUSTER_ACCESS_TOKEN="$(pass show community-tc/root)"
+   unset TASKCLUSTER_CERTIFICATE
+   tc-admin diff || true
+   tc-admin diff --ids-only || true
    echo
    echo 'Applying in 60 seconds (Ctrl-C to abort)....'
    echo
    sleep 60
    echo 'Applying!'
    echo
-   TASKCLUSTER_ROOT_URL=https://community-tc.services.mozilla.com tc-admin apply
+   tc-admin apply
    cd ../..
    rm -rf tc-admin
    ```


imagesets/README.md
```diff
@@ -147,14 +147,19 @@ complete:
...
+   TASKCLUSTER_ROOT_URL=https://community-tc.services.mozilla.com
+   TASKCLUSTER_CLIENT_ID=static/taskcluster/root
+   TASKCLUSTER_ACCESS_TOKEN="$(pass show community-tc/root)"
+   unset TASKCLUSTER_CERTIFICATE
+   tc-admin diff || true
+   tc-admin diff --ids-only || true
    echo
    echo 'Applying in 60 seconds (Ctrl-C to abort)....'
    echo
    sleep 60
```
lotas, line 159: What if we remove this sleep and just do a prompt instead? Would be more foolproof :)

```sh
read -p "Apply changes (yes/no)?" choice
case "$choice" in 
  yes ) tc-admin apply;;
  no ) echo "Ok ok 🐥 ";;
  * ) echo "invalid";;
esac

```

imagesets/README.md
```diff
@@ -147,14 +147,19 @@ complete:
    pip3 install -e .
    which tc-admin
    pass git pull
-   TASKCLUSTER_ROOT_URL=https://community-tc.services.mozilla.com tc-admin diff || true
+   TASKCLUSTER_ROOT_URL=https://community-tc.services.mozilla.com
+   TASKCLUSTER_CLIENT_ID=static/taskcluster/root
+   TASKCLUSTER_ACCESS_TOKEN="$(pass show community-tc/root)"
```
lotas, line 152: I think it needs to be truncated, it includes some extra lines

lotas:
```
access key
<blank line>
secret name
```

so head -n 1 ?

lotas: also, didn't work for me without `export` on those three.. maybe because tc-admin is creating some sub-shell or is being executed with those env vars not available to it

run-tc-admin.sh
```diff
@@ -0,0 +1,71 @@
...
+    read -p "Apply changes (yes/no)? " choice
+    case "$choice" in
+      yes)
+        echo
+        echo 'Applying!'
+        echo
+        tc-admin apply
+        done=true;;
+      no)
+        echo "Ok ok 🐥"
```
lotas, line 43: 😬 nice! Thanks

run-tc-admin.sh
```diff
@@ -0,0 +1,71 @@
...
+  which tc-admin
+  git clone ssh://gitolite3@git-internal.mozilla.org/taskcluster/secrets.git "${PASSWORD_STORE_DIR}"
+  TASKCLUSTER_ROOT_URL=https://community-tc.services.mozilla.com
+  TASKCLUSTER_CLIENT_ID=static/taskcluster/root
+  TASKCLUSTER_ACCESS_TOKEN="$(pass show community-tc/root | head -n 1)"
+  unset TASKCLUSTER_CERTIFICATE
+  tc-admin diff || true
+  tc-admin diff --ids-only || true
+  echo
+  echo 'Applying in 60 seconds (Ctrl-C to abort)....'
```
lotas, line 30: This will be misleading, let's drop it, as we show prompt below

PR #485, opened 2022-03-31
Add another fuzzing reducer pool for Android emulator crashes.

jschwartzentruber:

PR #488, opened 2022-04-06
Add missing fuzzing scopes.

jschwartzentruber:
Scopes for privileged docker were missed in #485. Required for Android emulator in Docker.

I've already applied these manually, so this should be a no-op.


lotas:
Oops :) will merge and check if reapply is necessary

PR #501, opened 2022-05-16
Bug 1768865: Refresh docker-worker *.taskcluster-worker.net certificate

petemoore:
For [bug 1768865](https://bugzil.la/1768865) we need to refresh the *.taskcluster-worker.net certificate before 16 June 2022. Only docker-worker image set is affected.


petemoore:
Already deployed and [confirmed working](https://bugzilla.mozilla.org/show_bug.cgi?id=1768865#c2).

matt-boris (approved):
Nice! Thanks Pete 🎉

PR #502, opened 2022-05-31
Add additional artifact path for bugmon hooks

pyoor:
Store entries from `/tmp/grizzly` as artifacts.

### DIFF
--- a/config/projects/fuzzing.yml
+++ b/config/projects/fuzzing.yml
@@ -307,6 +307,9 @@ fuzzing:
             project/fuzzing/bugmon:
               path: /bugmon-artifacts
               type: directory
+            project/fuzzing/grizzly:
+              path: /tmp/grizzly
+              type: directory
         metadata:
           name: bugmon
           description: Hook for triggering bugmon monitor tasks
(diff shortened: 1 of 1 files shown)


matt-boris (approved):
LGTM as long as it's fine storing in the `/tmp` directory? I don't see that being used anywhere else in the repo, so for consistency, I think `/grizzly-artifacts` may be better.

pyoor:
@matt-boris, sorry but I'm not sure I follow.  Are you referencing the fact that I'm creating artifacts for files in `/tmp/grizzly` or the location of the indexed artifact?  We're using similar paths in our fuzzing hooks.  See [here](https://community-tc.services.mozilla.com/tasks/d8JjEQo9RAGmabC3fGAIpA/definition).

matt-boris:
@pyoor I was referring to the former. Wasn't sure if it was a standard used or not, which is why I brought it up. Thanks, LGTM 👍🏻

matt-boris:
Applied!

```
Updating Hook=project-fuzzing/bugmon-confirm
Updating Hook=project-fuzzing/bugmon
```

PR #505, opened 2022-07-14
Add Windows 2016 Server Imageset with AMD drivers

pyoor:


petemoore (commented):
Jason and I chatted on matrix about this - we've decided we prefer to adapt the code as follows:

  * put static setup of IAM entities under the existing repository `/misc` directory (i.e. a script with `aws iam` commands together with any json files etc, to create e.g. a role called `S3ReadOnly`, that isn't specific to a particular image set)
  * have a file called `aws_run_instances_parameters` which replaces the current `aws_base_instance_type` file
  * this single file will contain all additional parameters to pass in to the [aws run-instances command](https://github.com/mozilla/community-tc-config/blob/df2c2b9d0277c07aca039cb79938af3b775d665c/imagesets/imageset.sh#L278), so it could contain e.g.

    ```
    --instance-type g4ad.xlarge --iam-instance-profile S3ReadOnly
    ```

  * an `aws_run_instances_parameters` file will exist in each imageset directory, as `aws_base_instance_type` does now (since `--instance-type` always needs to be specified)
  * [this line](https://github.com/mozilla/community-tc-config/blob/df2c2b9d0277c07aca039cb79938af3b775d665c/imagesets/imageset.sh#L278) will simply have `$(cat aws_run_instances_parameters)` where it currently has `--instance-type $(cat aws_base_instance_type)`
  * the existing `aws_base_instance_type` files from the existing imageset directories will be migrated to `aws_run_instances_parameters`, and have the text `--instance-type ` prepended to their content

imagesets/imageset.sh
```diff
@@ -274,8 +274,13 @@ function aws_update {
...
+    PROFILE="Name=$(cat aws_instance_profile)"
+  fi
+
   # Create new base AMI, and apply user-data filter output, to get instance ID.
-  if ! INSTANCE_ID="$(aws --region "${REGION}" ec2 run-instances --image-id "${AMI}" --key-name "${IMAGE_SET}_${REGION}" --security-groups "rdp-only" "ssh-only" --user-data "$(cat "${TEMP_SETUP_SCRIPT}")" --instance-type $(cat aws_base_instance_type) --block-device-mappings DeviceName=/dev/sda1,Ebs='{VolumeSize=75,DeleteOnTermination=true,VolumeType=gp2}' --instance-initiated-shutdown-behavior stop --client-token "${UNIQUE_NAME}" --query 'Instances[*].InstanceId' --output text 2>&1)"; then
+  if ! INSTANCE_ID="$(aws --region "${REGION}" ec2 run-instances --image-id "${AMI}" --key-name "${IMAGE_SET}_${REGION}" --security-groups "rdp-only" "ssh-only" --user-data "$(cat "${TEMP_SETUP_SCRIPT}")" --instance-type $(cat aws_base_instance_type) $INSTANCE_PROFILE_ARGS --block-device-mappings DeviceName=/dev/sda1,Ebs='{VolumeSize=75,DeleteOnTermination=true,VolumeType=gp2}' --instance-initiated-shutdown-behavior stop --client-token "${UNIQUE_NAME}" --query 'Instances[*].InstanceId' --output text ${PROFILE:+--iam-instance-profile $PROFILE} 2>&1)"; then
```
petemoore, line 283: I think `$INSTANCE_PROFILE_ARGS` can be removed here.

petemoore (commented):
Let's add a script called e.g. `/misc/aws_generate_entities.sh` that generates all the required entities, e.g. containing something like:

```
#!/usr/bin/env bash

set -eu
set -o pipefail

cd "$(dirname "${0}")"

./aws-create-instance-profile.sh \
  aws_ec2_generic_trust_policy.json \
  aws_s3_amd_driver_access_policy.json \
  EC2STSAssumeRole \
  S3ReadAMDWindows \
  AMDDriverS3Access
```

This way, we just have a single script that manages the creation of all the entities that are needed by any of the image sets.

imagesets/imageset.sh
pyoor: Good catch!

petemoore (approved):
This looks perfect, many thanks! :-)

PR #508, opened 2022-07-27
Add missing scopes required by covdiff

pyoor:
Apologies but it looks like I was also missing scopes to grant access to a required secret.


matt-boris:
Applied

```
Updating Role=hook-id:project-fuzzing/covdiff
```

PR #512, opened 2022-09-26
Use GPU instances for Grizzly reducer Windows pool.

jschwartzentruber:
We're seeing many task failures since enabling fuzzing on GPU instances because we later try to reduce those crashes on non-GPU instances.


lotas (approved):

lotas:
```
Updating WorkerPool=proj-fuzzing/grizzly-reduce-worker-windows
Updating Role=worker-pool:proj-fuzzing/grizzly-reduce-worker-windows
```

Changes applied!

PR #513, opened 2022-10-06
Add build-essential to ubuntu 18.04 and upgrade go and TC versions to latest

matt-boris:


imagesets/generic-worker-ubuntu-18-04/bootstrap.sh
```diff
@@ -33,7 +33,8 @@ start_time="$(date '+%s')"
 
 retry apt update
 DEBIAN_FRONTEND=noninteractive apt upgrade -yq
-retry apt install -y apt-transport-https ca-certificates curl software-properties-common gzip python3-venv python-virtualenv
+# build-essential is needed for running go test with the -vet=off flag as of go1.19
```
petemoore, line 36: I think(?) only if `-race` is also specified

petemoore (commented):
Thanks for this Boris!

I have some thoughts:

1) It might be that we can remove go in the base image, since our tasks install it. I wonder if the linking problem that we had was because of something that was getting picked up from the go1.18.1 host installation, that conflicted with the go1.19.2 installation of the task. We can talk about this, as there are arguments on both sides worth exploring.
2) Maybe we don't need build-essential, but just gcc - although maybe also nice to have a comprehensive set of tools by default. Again it is quite a nuanced issue, because on one hand it is nice to have reasonable defaults for installed tools, but on the other hand it encourages tasks not to declare dependencies and pin them, so system updates can break user tasks.
3) There is an update to a relman-specific image set, which probably isn't used by the taskcluster project. That is probably a good thing, we just need to make sure we let the team (Release Management - relman) know that we are patching their base images, and keep an eye on their tasks, to see if they start failing after this change.
4) For each of the bootstrap scripts that have been updated, we should build the new machine images (run `imageset.sh` script) so that the latest version in the repository is consistent with the live images. I think this is done for `generic-worker-ubuntu-20-04` but perhaps not for the other image sets yet?

Anyway this is a great start, and I'm really happy to see you were able to build new images successfully. Let's discuss the options for the above points when we next meet up.

Thanks Boris!

matt-boris:
I'm going to try removing the go install on the workers, as our tasks do this already. We'll need to be sure that other tasks in other worker pools aren't relying on the go binary to be installed on the worker.

This could've been the root issue we were having in upgrading to go1.19 - the task was installing go1.19, while the worker had go1.18 installed already - possibly some conflicts.

So I'll try removing the go install on the worker, as well as removing the new build-essential install and see what happens.

lotas (approved):

PR #520, opened 2022-10-27
Fix docker-worker config for disableSeccomp feature in proj-taskcluster/dw-ci worker pool

petemoore:
The docker-worker _config setting_ to say whether the worker accepts tasks that use the `disableSeccomp` feature is `allowDisableSeccomp`.


matt-boris (approved):
Great catch!

PR #526, opened 2022-10-31
docker-worker imageset split into docker-worker-2021 and docker-worker-2022

petemoore:
Since it may be beneficial to roll out the new docker worker image set gradually, I have introduced image sets docker-worker-2021 and docker-worker-2022. This way, some projects can remain on the 2021 docker-worker image sets while other projects adopt and test out the newer image sets.

CC @pyoor @jschwartzentruber


lotas (approved):
nice!

PR #530, opened 2022-11-18
Move bugmon-processor to gcp and bugmon-pernosco to aws

pyoor:
Although the audio/video issues have not been resolved on aws, the `bugmon-pernosco` pool must use aws (due to the bare-metal requirement).  This pool will not actually be enabled via the hook until the audio/video issues have been resolved.

Additionally, the `bugmon-processor` pool should use gcp.

### DIFF
--- a/config/projects/fuzzing.yml
+++ b/config/projects/fuzzing.yml
@@ -19,19 +19,18 @@ fuzzing:
       owner: jkratzer@mozilla.com
       emailOnError: false
       imageset: docker-worker-2022
-      cloud: aws
+      cloud: gcp
       minCapacity: 0
       maxCapacity: 10
       workerConfig:
         dockerConfig:
           allowPrivileged: true
           allowDisableSeccomp: true
-      securityGroup: docker-worker
     bugmon-pernosco:
       owner: jkratzer@mozilla.com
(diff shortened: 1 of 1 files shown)


config/projects/fuzzing.yml
```diff
@@ -19,7 +19,7 @@ fuzzing:
       owner: jkratzer@mozilla.com
       emailOnError: false
       imageset: docker-worker-2022
-      cloud: aws
+      cloud: gcp
```
matt-boris, line 22: Not sure if it'll affect this config or not, but I'd remove the `securityGroup: docker-worker` field here now, as that's used for AWS.

config/projects/fuzzing.yml
```diff
@@ -31,7 +31,7 @@ fuzzing:
       owner: jkratzer@mozilla.com
       emailOnError: false
       imageset: docker-worker-2022
-      cloud: gcp
+      cloud: aws
```
matt-boris, line 34: Now you should add this security group since it's moving over to AWS.

```suggestion
      cloud: aws
      securityGroup: docker-worker
```

matt-boris:
Applied!

PR #531, opened 2022-11-23
Add a role for testing #5777

djmitche:
This should allow me to
 * create tasks in a test task queue, for a specific projectId
 * create a worker polling that task queue

Depending on the configuration of the object service, this may or may not be able to create objects -- we'll see!


lotas:
@djmitche config was applied

PR #538, opened 2022-12-05
Remove unused project telemetry

petemoore:
In production, there are no tasks for this project, assuming it is dead.
Contact email for project (in worker pool definition) was for an individual, and that email address no longer works.


matt-boris (approved):
Ty 👍🏻

PR #539, opened 2022-12-06
Add allowDisableSeccomp to bugmon-monitor

pyoor:
This is required for running pernosco-submit on the bugmon reporter instances.


matt-boris:
Hey @pyoor, we're since removed this config after finding out privileged mode is sufficient.

https://github.com/taskcluster/taskcluster/pull/5800#issuecomment-1330766821

PR #540, opened 2022-12-06
chore(cleanup): remove stale `disableSeccomp` config

matt-boris:
Since https://github.com/taskcluster/taskcluster/pull/5800/files#diff-efc540af1977441446fbe043646cb2d22ee386c0a5d50a4d0ebafc1312ba061fL51-L54, this config is no longer used/needed.

### DIFF
--- a/config/projects/fuzzing.yml
+++ b/config/projects/fuzzing.yml
@@ -25,7 +25,6 @@ fuzzing:
       workerConfig:
         dockerConfig:
           allowPrivileged: true
-          allowDisableSeccomp: true
     bugmon-pernosco:
       owner: jkratzer@mozilla.com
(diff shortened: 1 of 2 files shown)


petemoore (approved):

petemoore:
Actually, let's hold off landing this for now as it looks like fuzzing currently require it, and are currently using docker-worker 44.23.0: https://community-tc.services.mozilla.com/tasks/Vcnm03yrT6WN0kaySSlCzQ/runs/0/logs/public/logs/live.log#L6

petemoore:
I'll close this for now so that we don't accidentally land it, but if we do decide we can do it later, we can always reopen it.

PR #545, opened 2022-12-14
Updated smoketest client name

lotas:
To align with existing envs: https://github.com/mozilla-services/cloudops-infra/pull/4571


petemoore (approved):
Looks good, but do we need to update anything else at the same time that references this name?

matt-boris (approved):

PR #535, opened 2022-12-02
Standardise on ubuntu 22.04

petemoore:
Move all ubuntu imagesets to ubuntu 22.04 LTS.

PR #546, opened 2022-12-14
fix(tc release): fix tc release worker pool to run tasks as current user (root)

matt-boris:
This was accidentally reverted in https://github.com/mozilla/community-tc-config/pull/535.


lotas (approved):

lotas (approved):
lgtm

PR #550, opened 2022-12-15
Add AMD drivers to the Windows 2022 imageset

jschwartzentruber:
We're already using the new Windows 2022 images for non-GPU instances. I believe this is all that's required to add the drivers to the imageset.

### DIFF
--- a/imagesets/generic-worker-win2022/aws_instance_profile
+++ b/imagesets/generic-worker-win2022/aws_instance_profile
@@ -1,0 +1,1 @@
+AMDDriverS3Access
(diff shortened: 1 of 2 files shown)


jschwartzentruber:
On further testing, this script fails to install the driver. The [AWS AMD driver install documentation](https://docs.aws.amazon.com/AWSEC2/latest/WindowsGuide/install-amd-driver.html#download-amd-driver) states:

> The AMD driver is only supported for Windows Server 2016 and Windows Server 2019 operating systems.

petemoore:
Oh wow. And the docs don't even say why it is only supported on those platforms, or when it is expected to be supported, or what the alternatives are. :-/

PR #553, opened 2023-01-10
New roles would be assumed for releases

lotas:
Due to the changes in github service of Taskcluster, we need to adjust few roles here.

There are just two projects who defined separate `:release` roles, so in order to keep it working we'd need to merge this one right after we make a release to community.

Please see https://github.com/taskcluster/taskcluster/pull/5940

### DIFF
--- a/config/projects/firefoxreality.yml
+++ b/config/projects/firefoxreality.yml
@@ -53,6 +53,7 @@ firefoxreality:
         - secrets:get:project/firefoxreality/fr/release-signing-token
       to:
         - repo:github.com/MozillaReality/FirefoxReality:release
+        - repo:github.com/MozillaReality/FirefoxReality:release:*
     - grant:
         - queue:route:index.project.firefoxreality.*
       to:
--- a/config/projects/git-cinnabar.yml
+++ b/config/projects/git-cinnabar.yml
@@ -47,5 +47,6 @@ git-cinnabar:
         - queue:route:index.project.git-cinnabar.*
       to:
         - repo:github.com/glandium/git-cinnabar:release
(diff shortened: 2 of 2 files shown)


matt-boris (dismissed):
Changes LGTM, but we could also just _add_ this new line instead and then go back with a subsequent PR to clean up the unused once everything's been tested successfully.

lotas:
yeah, I guess we could keep both, true. This is also an option. Either we keep this one unmerged or we merge extra role and have a PR ready to remove one after. I'll change this

config/projects/firefoxreality.yml
```diff
@@ -53,6 +53,7 @@ firefoxreality:
         - secrets:get:project/firefoxreality/fr/release-signing-token
       to:
         - repo:github.com/MozillaReality/FirefoxReality:release
+        - repo:github.com/MozillaReality/FirefoxReality:release:*
```
matt-boris, line 56: Or maybe even `...release*`?

matt-boris: Then we wouldn't need a subsequent PR at all.

matt-boris (approved):

matt-boris:
```bash
Creating Role=repo:github.com/glandium/git-cinnabar:release:*
Creating Role=repo:github.com/MozillaReality/FirefoxReality:release:*
```

jamesjames1122 (commented):
gb

config/projects/firefoxreality.yml
jamesjames1122: @jamesjames1122

PR #556, opened 2023-01-25
Enable sshd on proj-taskcluster windows workers

petemoore:
It might be easiest to review the commits in sequence. This PR fixes the installation of an ssh daemon on the Windows workers (for everyone) but only applies the ssh security group to the proj-taskcluster Windows workers running in AWS. This should make it possible to ssh onto Windows workers when there are problems. It was already possible to RDP onto them.

Note, this required a fix to the python code to allow workers to be in more than one security group.

Thanks!

### DIFF
--- a/config/aws.yml
+++ b/config/aws.yml
@@ -47,18 +47,22 @@ security_groups:
     no-inbound: sg-00c4014bc978171d5
     docker-worker: sg-0d2ff88f36a05b499
     rdp: sg-0ddc5eae2e56a43c5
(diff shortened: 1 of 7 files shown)


lotas (approved):
awesome!

i wonder if there would be a safer approach, where we don't add this security group by default, but only when needed. Not sure if this is something that worker-manager should be responsible for, but if we add an API endpoint, then we could already enable this group though TC UI (I assume AWS SDK allows to apply new groups)

upd: ah, I see, there is a key-only authentication, should be fine then 👍

PR #557, opened 2023-02-02
Bump tc version to latest and update passwordstore location

matt-boris:
~~`gcc` appears to be required to fix the CI failure from https://github.com/taskcluster/taskcluster/pull/6050.~~
From https://go.dev/doc/go1.20:
>On Linux and other Unix systems, and on Windows, a host C toolchain is required to use the race detector.

Edit: I'll try to fix this in a later PR.

This change also clones the passwordstore from its new repo. Please follow these instructions to get it to work for you @petemoore and @lotas (@hwine too if you'd like to set it up).
```bash
  # Register your ssh public key with https://source.cloud.google.com/user/ssh_keys?register=true
  #
  # Add the following to your ~/.ssh/config:
  #
  # Host source.developers.google.com
  #  User <user>@mozilla.com
  #  UpdateHostKeys yes
  #  IdentityFile <path to your private key>
  #  Port 2022
  #
```

View updated readme [here](https://source.cloud.google.com/taskcluster-passwords/secrets).

### DIFF
--- a/imagesets/generic-worker-ubuntu-22-04-staging/bootstrap.sh
+++ b/imagesets/generic-worker-ubuntu-22-04-staging/bootstrap.sh
@@ -4,7 +4,7 @@ set -exv
 exec &> /var/log/bootstrap.log
 
 # Version numbers ####################
-TASKCLUSTER_REF='v46.0.0'
+TASKCLUSTER_REF='v47.0.3'
 ######################################
 
 function retry {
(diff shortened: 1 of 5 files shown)


lotas:
that's really nice! I didn't know gcp has this service, just switched over, works fine! Thanks

lotas (approved):

PR #558, opened 2023-02-08
Bump tc-admin and json-e to latest versions

matt-boris:


petemoore (approved):
Thanks Boris!

Should we integrate dependabot / renovate here? I wonder what changed in json-e and tc-admin between these releases, I don't think we've touched them for a while. Maybe the previous versions were ancient?

matt-boris:
> Should we integrate dependabot / renovate here?

I'll add it now in a new PR 👍🏻

PR #563, opened 2023-03-16
update instance type availability in gce/ec2

petemoore:
Updated the instance type availability databases for gce/ec2.


matt-boris (approved):
Thanks! 🎉

PR #570, opened 2023-04-17
Add scopes for fuzzmanager-config repo

jschwartzentruber:

PR #571, opened 2023-04-18
Fix typo in fuzzing role.

jschwartzentruber:
Sorry @matt-boris , I didn't notice this in #570 until it was applied.


matt-boris (approved):
Np 👍🏻

PR #549, opened 2022-12-15
Restore proj-fuzzing/grizzly-reduce-worker-windows

petemoore:
Sorry about this @jschwartzentruber @pyoor !

This should restore your previous proj-fuzzing/grizzly-reduce-worker-windows worker pool machine image.

PR #574, opened 2023-05-02
Update all worker pools to generic-worker 49.1.2

petemoore:
This is useful for testing https://bugzilla.mozilla.org/show_bug.cgi?id=1815711 on Windows, and keeps everything else up-to-date at the same time.

### DIFF
--- a/imagesets/generic-worker-win2016-amd/bootstrap.ps1
+++ b/imagesets/generic-worker-win2016-amd/bootstrap.ps1
@@ -1,3 +1,10 @@
+# This script is not maintained!!
+#
+# See https://github.com/mozilla/community-tc-config/pull/549#issuecomment-1353426207
+# This is just a working version used by the fuzzing team, until they are able
+# to migrate to the standard Windows Server image set that we provide.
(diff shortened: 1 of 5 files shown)


imagesets/generic-worker-win2016-amd/bootstrap.ps1
```diff
@@ -1,4 +1,4 @@
-$TASKCLUSTER_VERSION = "44.21.0"
+$TASKCLUSTER_VERSION = "49.1.2"
```
matt-boris, line 1: If I remember correctly, this is pinned on this version for a reason. We should look into it before merging.

petemoore: ah indeed - found it!

https://github.com/mozilla/community-tc-config/pull/549#issuecomment-1353426207

i'll add a comment so we don't forget again....

matt-boris (approved):
Perfect, thanks! Thanks for adding in the note on that pinned version too 🎉

PR #584, opened 2023-05-22
Disable AV and disk indexing in Windows images for improved performance

marco-c:
I think this is valuable in general, so applying to both the generic image and the relman specific one.


matt-boris (approved):

lotas (approved):

petemoore (approved):
Great, many thanks!

PR #586, opened 2023-05-24
Migrate relman machine image from Windows Server 2012 R2 to Windows Server 2022

petemoore:
Note, if this works, and fixes the issue, we should rename the image set directory and the worker pool name to reflect the updated version of Windows. But since this might not work, haven't done that yet (will affect the .taskcluster.yml files that specify the worker pool too).

@marco-c Can you take a look and let me know what you think?


marco-c (approved):
Llgtm, I can trigger the task to make sure it works after you apply this and then we can rename

lotas (approved):

PR #587, opened 2023-05-24
Rename proj-relman/win2012r2 to proj-relman/win2022

petemoore:
Release management Windows images are now based on Windows Server 2022 rather than Windows Server 2012 R2.


marco-c (approved):

petemoore:
Applied!

PR #589, opened 2023-05-25
Configure locations of livelog and taskcluser-proxy on relman-win2022

petemoore:
This should fix livelog on this imageset.

CC @marco-c


petemoore:
(applied already, and has fixed the issue)

marco-c (approved):
\o/

PR #595, opened 2023-06-08
exclude audit client

lotas:
That's the client used to run periodic audit reports


petemoore:
I think we can just define the client, rather than manage it externally?

petemoore (dismissed):
Note, for me it is ok if you want to run tc-admin from this branch directly, and modify the list of scopes until it works, rather than needing to create a new PR for each iteration of fixes adding a scope at a time (if that makes sense).

lotas:
added scheduler scope, works now

matt-boris (approved):

PR #597, opened 2023-06-14
Install v4l2loopback for gcp

matt-boris:
AWS has this kernel module by default. We will then, in generic worker, run:

```bash
modprobe v4l2loopback video_nr=0
chmod 660 /dev/video0
chown root:<taskUser> /dev/video0
```

so that the `/dev/video0` device is available to the task user.


petemoore (approved):
Nice!

PR #599, opened 2023-06-19
Enable nested virtualization for wpt project workers

jgraham:


petemoore (approved):
Looks great! I will land this. Probably we should have the same for aws/azure, and update the docs to explain the new launchConfig setting, but this is certainly landable, and the `tc-admin diff` generates the intended result. Thanks @jgraham for implementing and @jcristau for support!

PR #601, opened 2023-06-21
Apply staging ubuntu 22.04 improvements to production (where effective)

petemoore:
We had some patches applied to staging Ubuntu 22.04 image set that were effective and would be useful for production.
Note, new image sets do not need to be built; this change simply improves reliability of the image building process when there are failures downloading release artifacts of taskcluster. But no failures occurred in the last round of image building image sets, so no new image sets are needed.

### DIFF
--- a/imagesets/generic-worker-ubuntu-22-04/bootstrap.sh
+++ b/imagesets/generic-worker-ubuntu-22-04/bootstrap.sh
@@ -53,16 +53,16 @@ if [[ "%MY_CLOUD%" == "google" ]]; then
 fi
 
 cd /usr/local/bin
-retry curl -L "https://github.com/taskcluster/taskcluster/releases/download/${TASKCLUSTER_VERSION}/generic-worker-multiuser-linux-amd64" > generic-worker
-retry curl -L "https://github.com/taskcluster/taskcluster/releases/download/${TASKCLUSTER_VERSION}/start-worker-linux-amd64" > start-worker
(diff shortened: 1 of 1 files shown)


imagesets/generic-worker-ubuntu-22-04/bootstrap.sh
```diff
@@ -53,16 +53,16 @@ if [[ "%MY_CLOUD%" == "google" ]]; then
...
+retry curl -fsSL "https://github.com/taskcluster/taskcluster/releases/download/${TASKCLUSTER_VERSION}/livelog-linux-amd64" > livelog
+retry curl -fsSL "https://github.com/taskcluster/taskcluster/releases/download/${TASKCLUSTER_VERSION}/taskcluster-proxy-linux-amd64" > taskcluster-proxy
 chmod a+x generic-worker start-worker taskcluster-proxy livelog
 
 mkdir -p /etc/generic-worker
 mkdir -p /var/local/generic-worker
-./generic-worker --version
-./generic-worker new-ed25519-keypair --file /etc/generic-worker/ed25519_key
+/usr/local/bin/generic-worker --version
+/usr/local/bin/generic-worker new-ed25519-keypair --file /etc/generic-worker/ed25519_key
```
petemoore, line 65: The primary reason for this change is to reduce the diff between the staging version and the production version. Of the two, I preferred the staging version of these lines, as absolute path is more explicit, so I updated production to match staging, rather than changing staging to match production.

matt-boris (approved):
Nice, thanks Peeeeeeeeeeeeeet 👍🏻

PR #602, opened 2023-06-23
increase max capacity of gw-ubuntu-22-04 pool to 20

matt-boris:
Now that we're shifting more CI tasks to this pool, I think it's appropriate to increase the max capacity to 20 for now.


lotas (approved):
👍

PR #608, opened 2023-06-27
chore: bump python versions

matt-boris:


.taskcluster.yml
```diff
@@ -50,8 +50,8 @@ tasks:
               git config advice.detachedHead false &&
               git checkout ${repo.ref} &&
               pip install black &&
-              black --check generate
-        - image: 'python:3.6'
+              black --diff generate
```
matt-boris, line 53: `--diff` will output what `black` _would have_ formatted. Similar to `--check` except for the fact that it outputs a more productive message if there are any issues.

lotas (approved):
:+1:

PR #612, opened 2023-07-13
Add FreeBSD worker pool for testing generic-worker and client/worker tools

petemoore:
In taskcluster/taskcluster#6399 we are fixing FreeBSD issues in Generic Worker. This PR is about setting up a worker pool for that work so testing on FreeBSD is part of the taskcluster CI.


config/projects/taskcluster.yml
```diff
@@ -82,6 +82,14 @@ taskcluster:
       # note: workerConfig not specified here as configuration
       # is stored directly on manually provisioned worker
 
+    gw-ci-freebsd:
+      owner: taskcluster-notifications+workers@mozilla.com
+      emailOnError: true
+      imageset: generic-worker
+      cloud: static
+      lifecycle:
+        reregistrationTimeout: 2592000 # 30 days
```
matt-boris, line 91:
```suggestion
      lifecycle:
        reregistrationTimeout: 2592000 # 30 days
      workerConfig:
        genericWorker:
          config:
            runTasksAsCurrentUser: true
```

Don't you need this set if we're calling this a `gw-ci-*` pool?

petemoore: static pools don't fetch their config from worker manager 😬 

they probably should, but they don't 😬

matt-boris: Ah sorry I didn't realize this was static!

matt-boris (approved):

PR #628, opened 2023-07-14
minor doc fix

petemoore:
Pretty trivial fix - just wanted to highlight that this worker pool isn't for _all_ CI tasks, only for the Generic Worker CI tasks. It could have looked like it was for all CI tasks, but not other proj-taskcluster tasks, which isn't the case.


lotas (approved):
Thanks!

PR #629, opened 2023-07-14
Initial imageset structure for freebsd

petemoore:
This is needed for the first time we run imageset.sh against freebsd. The structure needs to be here, so that it can update the AMIs etc.


matt-boris (approved):
👍🏻

PR #633, opened 2023-07-18
chore: switch default kernel on aws to install snd-aloop module

matt-boris:
Required for https://github.com/taskcluster/taskcluster/pull/6419.


petemoore (approved):
Love it!

PR #634, opened 2023-07-18
proj-tascluster generic-worker staging pool also for google/gcp

petemoore:
Note, this renames the existing staging pool in AWS, but I think that is ok, because I don't think anything references it.

Now we'll have two pools, and then worker pool name will show whether it is running in aws or google/gcp.


matt-boris (approved):
Nice, ty!

PR #663, opened 2023-08-09
Create initial imageset generic-worker-ubuntu-22-04-arm64

### LINKED ISSUE #662: Create Linux/aarch64 GCP worker images
Originating from:
  * https://bugzilla.mozilla.org/show_bug.cgi?id=1677963 and
  * https://mozilla-hub.atlassian.net/browse/RELOPS-265

Unfortunately this was stuck in the wrong team's queue since November 2020! Found out about this last night.

petemoore:
This initial PR doesn't create any concept of process architecture for an image set, instead it duplicates the existing amd64 image set `generic-worker-ubuntu-22-04` and calls it `generic-worker-ubuntu-22-04-arm64` and sets the machine type for image building to `t2a-standard-4` so that it effectively is an arm64 image set.

In future we may reduce this duplication here, and make the image building machinery process-architecture-aware so that the an image set may be built for multiple process architectures. But this is the first step to unblocking issue #662.

Fixes #662.


petemoore:
Success! https://community-tc.services.mozilla.com/tasks/L9Hor4QiQ7KdwN4FeyiINA/runs/0/logs/live/public/logs/live.log

petemoore:
Both matt-boris and lotas are on PTO this week, but since the task ran above, I'm going to consider this a success and land it.

PR #664, opened 2023-08-12
Add kvm.enable_vmware_backdoor modprobe option for Ubuntu x86-64.

jschwartzentruber:
This is not as scary as it sounds. This enables a VMware compatible interface in kvm, which is [needed for some fuzzing tasks](https://github.com/AFLplusplus/AFLplusplus/blob/stable/nyx_mode/README.md#fuzzing-with-nyx-mode).

Some very low-level explanation of the feature here: https://www.spinics.net/lists/kvm/msg160883.html

### DIFF
--- a/imagesets/generic-worker-ubuntu-22-04/bootstrap.sh
+++ b/imagesets/generic-worker-ubuntu-22-04/bootstrap.sh
@@ -58,6 +58,12 @@ retry apt-get update
 retry apt-get install -y docker-ce docker-ce-cli containerd.io
 retry docker run hello-world
 
+# configure kvm vmware backdoor
+# this enables a vmware compatible interface for kvm, and is needed for some fuzzing tasks
+cat > /etc/modprobe.d/kvm-backdoor.conf << "EOF"
+options kvm enable_vmware_backdoor=y
+EOF
+
 cd /usr/local/bin
 retry curl -fsSL "https://github.com/taskcluster/taskcluster/releases/download/${TASKCLUSTER_VERSION}/generic-worker-multiuser-linux-${ARCH}" > generic-worker
 retry curl -fsSL "https://github.com/taskcluster/taskcluster/releases/download/${TASKCLUSTER_VERSION}/start-worker-linux-${ARCH}" > start-worker


petemoore:
Ha, indeed that does _sound_ scary. I didn't really understand the details.

@hwine Do you have any concerns/thoughts about this?

hwine:
A couple of questions, as from what Pete has told me, there's no way to make this a "fuzzing job only" change. It will affect all images for all users.

I'm not an expert, so did a bit of searching. [This VMware forum article](https://communities.vmware.com/t5/ESXi-Discussions/What-is-the-virtual-machine-backdoor/td-p/2128896) is particularly informative (and appears to be from an employee of VMWare):
- "Backdoor" is a poor choice of feature name - it's more of a well-known-path-for-hypervisor-comms.
- It _appears_ to be unique to VMWare hypervisors (or at least that's it's intended scope - _I think_).
- [This comment](https://communities.vmware.com/t5/ESXi-Discussions/What-is-the-virtual-machine-backdoor/m-p/2128901/highlight/true#M197956) highlights there is room for abuse of the comms channel.

Since it is a back channel, somewhat related to nesting hypervisors, I tend to be conservative. We've had a history in TC of being concerned about excessive privileges being available to workers running untrusted code.

@jschwartzentruber - can you speak to
- any risks you see to non fuzzing jobs related to enabling this mode?
  - if so, do you see ways to mitigate those risks?
- is the alternate enabling method (using the `iopl()` system call, as mentioned [here](https://communities.vmware.com/t5/ESXi-Discussions/What-is-the-virtual-machine-backdoor/m-p/2128903/highlight/true#M197958)) sufficient for your use case?

With that information, I expect the next step to either be approval, or move to an RRA.

jschwartzentruber:
Thanks for looking into this. I am not an expert here, but from my reading of it, the feature in question (`kvm.enable_vmware_backdoor`) is about allowing the same backdoor from the VMware posts above, but in the KVM hypervisor rather than ESXi. The same comments about risk would apply though. The relevant comment from the post you linked is:

> there is always the possibility that one of the hypercall implementations enables some kind of security exploit

My read of this is that the risk is of potentially vulnerable code in the hypercall handler allowing hypervisor escape from untrusted *virtualized* code. I don't think we're talking about an increased risk at the taskcluster host level, but I could be wrong. I am assuming the hypercall handlers we're talking about are run in userspace in QEMU, so any escape would compromise the QEMU process only.

I don't know if the alternative `iopl()` system call would work or not, this is an externally maintained fuzzing project.

petemoore (approved):

petemoore:
@jschwartzentruber This is now deployed in the community cluster. Let us know how you get on with the change, and if it works as expected. Thanks!

PR #667, opened 2023-08-15
Align the Generic Worker Ubuntu 22.04 image sets where they have diverged

petemoore:
Note, this also replaces imagesets/generic-worker-ubuntu-22-04-arm64/bootstrap.sh with a symbolic link to imagesets/generic-worker-ubuntu-22-04/bootstrap.sh.

Until now I've generally avoided having symbolic links inside git repositories, not being sure how well they work across OSes and file system implementations. But then I remembered, it is 2023. Surely it must be ok these days. And this is certainly a way to find out.

A possible drawback is that someone might patch e.g. `imagesets/generic-worker-ubuntu-22-04-arm64/bootstrap.sh` and think they are only updating the arm64 variant, not realising they are also changing `imagesets/generic-worker-ubuntu-22-04/bootstrap.sh` at the same time. But I _think_ the advantage of them keeping in sync still outweighs the disadvantage that someone could accidentally update both, thinking they are updating only one.

Note, another alternative strategy exists, which is to rework the configuration, so you can point to a bootstrap script that lives outside the imageset directory, but that is kind of what a symbolic link does anyway, just perhaps a little more covertly than would be ideal.


matt-boris (approved):
👍🏻

PR #671, opened 2023-08-31
Switch to spot instances in GCP.

jschwartzentruber:
Spot instances in GCP are documented here: https://cloud.google.com/compute/docs/instances/create-use-spot#api

We've been using these in fuzzing pools for some time without issue, although I have no access to the GCP console to confirm we're really getting spot instances.


petemoore (approved):
Thanks so much for this @jschwartzentruber! Especially for doing it in workers.py so that all the pools benefit. ❤️

PR #679, opened 2023-09-20
chore: remove unneeded (likely?) worker config

matt-boris:
@marco-c I think this is safe to remove for now. Does this look fine to you? The tasks will still have root access from _within_ the container with podman. Please let us know if you run into any issues without this config.


marco-c:
It should be safe to remove once https://github.com/mozilla/bugbug/pull/3668 lands and is deployed.

PR #682, opened 2023-09-28
Issue 681 - completely uninstall Windows Defender on Windows Server 2022

### LINKED ISSUE #681: Windows Defender interfering with Generic Worker unit test TestAbortAfterMaxRunTime
We are getting permanent failure of Generic Worker unit test TestAbortAfterMaxRunTime on Windows Server 2022.

Example: https://github.com/taskcluster/taskcluster/runs/17224277278

WinDefend service should not be running on Windows, but we can see from installation logs, that disabling it is no longer working:

```
Set-Service : Service 'Microsoft Defender Antivirus Service (WinDefend)' cannot be configured due 
to the following error: Access is denied
At 
C:\Windows\system32\config\systemprofile\AppData\Local\Temp\EC2Launch3594172634\UserScript.ps1:28 
char:1
+ Set-Service "WinDefend" -StartupType Disabled -Status Stopped
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : PermissionDenied: (System.ServiceProcess.ServiceController:ServiceCo 
   ntroller) [Set-Service], ServiceCommandException
    + FullyQualifiedErrorId : CouldNotSetService,Microsoft.PowerShell.Commands.SetServiceCommand
```

This comes from https://community.taskcluster-artifacts.net/BMIpNvPWTxSdBi7vWJChBA/0/public/logs/err.tmp which is an installation log from when the machine image was created.

We need to find a new way to disable Windows Defender, since the current mechanism is ineffective.

petemoore:
This is the command that RelOps suggested using.

Fixes #681

### DIFF
--- a/imagesets/generic-worker-win2022/bootstrap.ps1
+++ b/imagesets/generic-worker-win2022/bootstrap.ps1
@@ -24,8 +24,10 @@ function Expand-ZIPFile($file, $destination, $url)
 # allow powershell scripts to run
 Set-ExecutionPolicy Unrestricted -Force -Scope Process
 
-# Disable AV for IO speed
-Set-Service "WinDefend" -StartupType Disabled -Status Stopped
+# Issue 681: Uninstall Windows Defender as it can interfere with tasks,
+# degrade their performance, and e.g. prevents Generic Worker unit test
+# TestAbortAfterMaxRunTime from running as intended.
+Uninstall-WindowsFeature -Name Windows-Defender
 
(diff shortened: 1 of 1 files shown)


matt-boris (approved):
Changes LGTM, let's test first before merging and rolling out to all win pools 👍🏻

PR #685, opened 2023-10-10
Bump proj-relman/generic-worker-ubuntu-22-04 worker's max capacity to 50

marco-c:
We are using it in place of 'ci' in code-coverage and code-review, and 'ci' had 50 as max capacity.


lotas (approved):
Let's try :)

PR #687, opened 2023-10-10
Disable SysMain (Superfetch) to improve performance

tysmith:
This service is meant to improve performance over time. In the short term it consumes compute resources. This is not helpful for fuzzing and likely any other use case.


petemoore (approved):
TIL, many thanks!
Sorry for slow review, for some reason it didn't show up in my queue. Will try to fix!

PR #689, opened 2023-10-11
Fix pipelines

lotas:
Updated .taskcluster.yml provisioner to allow `proj-taskcluster/ci` workers to pick up tasks (instead of `proj-misc/ci`)

Updates community-tc-config hook after migration


matt-boris (approved):
Nice, thank you!

PR #693, opened 2023-10-25
Remove Defender from generic-worker-win2016-amd imageset

jschwartzentruber:


jschwartzentruber:
Close in favour of #694

PR #694, opened 2023-10-25
Update Windows Server 2016 (amd64) image for fuzzing workers

petemoore:
The AMD driver required by Windows fuzzing workers is still not supported on Windows Server 2022, so refreshing Windows Server 2016 image, to realign with Windows Server 2022.

This also unblocks an issue where fuzzing jobs need to run as root to disable/uninstall Windows Defender, since this has now been removed from the base image, like it is on Windows Server 2022 machine images. This means those fuzzing tasks will no longer need to run as Administrator, which should help with an issue where some tasks seem to inadvertently shutdown the machine, which causes tasks to be resolved as exception/claim-expired. When not running as Administrator, they should (hopefully) not be able to shutdown the machine.

### DIFF
--- a/imagesets/generic-worker-win2016-amd/bootstrap.ps1
+++ b/imagesets/generic-worker-win2016-amd/bootstrap.ps1
@@ -1,11 +1,12 @@
-# This script is not maintained!!
+# We still support Windows Server 2016 since the Fuzzing team depend on AMD drivers
+# which are not (yet) compatible with Windows Server 2022.
 #
-# See https://github.com/taskcluster/community-tc-config/pull/549#issuecomment-1353426207
-# This is just a working version used by the fuzzing team, until they are able
-# to migrate to the standard Windows Server image set that we provide.
-#
-# Do not update unless necessary!
-$TASKCLUSTER_VERSION = "44.21.0"
+# This image should only be used for this purpose - all other Windows workloads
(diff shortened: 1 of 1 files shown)


imagesets/generic-worker-win2016-amd/bootstrap.ps1
```diff
@@ -5,7 +5,7 @@
 # to migrate to the standard Windows Server image set that we provide.
 #
 # Do not update unless necessary!
-$TASKCLUSTER_VERSION = "44.21.0"
```
matt-boris, line 8: I thought this was pinned here for some reason. `disableSeccomp` stuff maybe? I don't quite remember.

jschwartzentruber: This comment was added here: https://github.com/taskcluster/community-tc-config/pull/574#discussion_r1182494971 but the justification linked there doesn't seem to indicate that TC should be pinned.

matt-boris: Ok, please just let us know if the newer generic worker version causes unexpected issues then.

@petemoore would you please also remove the comments above that line?

matt-boris (approved):

matt-boris (approved):

PR #702, opened 2023-12-06
Update taskcluster.yml

matt-boris:
Enable display device in staging launch config for gw-ubuntu-staging-google worker pool


petemoore (approved):
Thank you so much!!

PR #709, opened 2023-12-11
Switch from simpledrm to vkms to work on both arm64 and x86_64

petemoore:
Kudos to @jcristau for suggesting this fix! 🥳 

Note, I tested on a manually launched worker, and it worked perfectly.

PR #710, opened 2023-12-11
Install Linux extra modules and configure vKMS for all cloud providers

petemoore:
Until now we've only configured vKMS on Linux workers if they are running GCP, not in other clouds.
Testing AWS staging Ubuntu using the same technique as used with GCP in #709 in this PR.

Currently AWS staging [is using](https://community-tc.services.mozilla.com/tasks/WRw6TLQwT1-bHycctYV7Aw/runs/0/logs/public/logs/live.log#L34) an X11 graphical backend rather than DRM.

### DIFF
--- a/config/imagesets.yml
+++ b/config/imagesets.yml
@@ -133,13 +133,13 @@ generic-worker-ubuntu-22-04-staging:
         workerTypeMetadata:
           machine-setup:
             maintainer: taskcluster-notifications+workers@mozilla.com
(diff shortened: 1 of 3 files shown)


petemoore:
Testing in https://community-tc.services.mozilla.com/tasks/CDndA3ZURW-RWgVReR-OmA

petemoore:
Woooooooooo

* https://community-tc.services.mozilla.com/tasks/CDndA3ZURW-RWgVReR-OmA/runs/0/logs/public/logs/live.log#L34

lotas (approved):

PR #717, opened 2024-01-25
Add queue:pending-list:* and queue:claimed-list:* to anonymous role

petemoore:
No reason to hide this information, tasks are published on exchanges so the data could be scraped anyway.


matt-boris (approved):
👍🏻

matt-boris:
Applied!

```
Updating Role=anonymous
Updating Role=project-admin:*
```

PR #720, opened 2024-02-14
Create unix group snap_sudo on ubuntu workers for running snap commands in tasks

petemoore:
Will initially deploy to staging to test that it is working as expected.


petemoore:
Testing here: https://community-tc.services.mozilla.com/tasks/V78sZZBsTZi1bffY95jZ4A

petemoore:
It worked!

PR #552, opened 2022-12-19
Set bugmon worker pools to use docker-worker-legacy imageset

pyoor:

PR #721, opened 2024-02-21
Migrate bugmon from Docker Worker to Generic Worker

petemoore:
@pyoor Generic Worker can now run Docker Worker task payloads, so this _should_ work. It looks like these were set to legacy version of Docker Worker in #552 but the reason for the change isn't clear from the PR conversation, so I'm not sure what the problems were with the regular Docker Worker images.

In any case, Generic Worker should emulate all of the Docker Worker features, so hopefully things should Just Work™ with this change.

@matt-boris Do you remember why legacy was required over the standard images?

### DIFF
--- a/config/projects/fuzzing.yml
+++ b/config/projects/fuzzing.yml
@@ -8,25 +8,17 @@ fuzzing:
     bugmon-monitor:
       owner: jkratzer@mozilla.com
       emailOnError: false
-      imageset: docker-worker-legacy
+      imageset: generic-worker-ubuntu-22-04
       cloud: gcp
       minCapacity: 0
       maxCapacity: 5
-      workerConfig:
-        dockerConfig:
-          allowPrivileged: true
-          allowDisableSeccomp: true
     bugmon-processor:
       owner: jkratzer@mozilla.com
       emailOnError: false
-      imageset: docker-worker-legacy
+      imageset: generic-worker-ubuntu-22-04
       cloud: gcp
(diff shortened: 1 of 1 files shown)


matt-boris:
I thought it was something to do with them needing that `disableSeccomp` feature. We added it, then they said they didn't need it, so we removed it, but then I think they realized they still needed it. So they pinned the version to the docker worker that had it.

pyoor:
Setting `disableSeccomp` is required for us to run [rr](https://github.com/rr-debugger/rr/wiki/Docker) in the containers.

petemoore:
Working on support in d2g for this here: https://github.com/taskcluster/taskcluster/pull/6846

matt-boris (dismissed):

lotas (dismissed):

lotas (dismissed):

petemoore:
I've reduced the scope to just bugmon-monitor and bugmon-processor since bugmon-pernosco requires AWS and snd-aloop, and at the moment that isn't working for us.

We removed code in #635 which attempted to add support, but did not work. It looks like https://github.com/mozilla-platform-ops/monopacker/blob/main/scripts/ubuntu-bionic/11-snd-aloop.sh may provide some working code, since this is probably how support was added to the current working docker worker image. However, presumably we tried following this implementation previously, so it may be that there are other issues at play.

lotas (approved):

PR #723, opened 2024-03-05
Switch proj-fuzzing/ci to generic-worker

jschwartzentruber:
This will make ci-gw unnecessary, but we'll have to migrate tasks back to ci before removing it.

### DIFF
--- a/config/projects/fuzzing.yml
+++ b/config/projects/fuzzing.yml
@@ -47,13 +47,10 @@ fuzzing:
     ci:
       owner: fuzzing+taskcluster@mozilla.com
       emailOnError: false
-      imageset: docker-worker
+      imageset: generic-worker-ubuntu-22-04
       cloud: gcp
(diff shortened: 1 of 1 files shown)


petemoore (approved):
Thanks Jesse! Are there timing issues around landing this? When would you like us to do it? Another option is for us to run tc-admin from this branch, and check all is ok, before merging. Let us know how you would prefer to proceed! If you have no preference, I think I would prefer to run tc-admin first, check things are ok, and then only merge if all is good, so we don't need to revert with another PR etc.

jschwartzentruber:
This should be good to go. orion is the most complicated user of this pool and it is using ci-gw, everything remaining on ci should be simple dw tasks.

matt-boris:
@jschwartzentruber sorry this fell through the cracks for a bit, but I just merged and applied these changes 👍🏻 Please let us know if you have any issues.

PR #724, opened 2024-03-17
Change machine type for git-cinnabar linux workers

glandium:
I'm having some OOM issues with n2-standard-4, let's try with n2-highmem-4, which has the same number of vCPUs, but twice the amount of memory.


lotas (approved):
👍

lotas:
@glandium should be updated now

glandium:
It worked, thanks.

PR #729, opened 2024-05-16
Generate config/azure-machine-type-offerings.json in consistent order

petemoore:
This helps to keep diffs small when we regenerate `config/azure-machine-type-offerings.json`.


misc/update-azure-machine-types.sh
```diff
@@ -15,11 +15,11 @@
 cd "$(dirname "${0}")"
 
 rm '../config/azure-machine-type-offerings.json'
-locations=$(az account list-locations --query="[].name" --output tsv)
+locations=$(az account list-locations --query="[].name" --output tsv | sort -u)
 
 list="["
 for location in $locations; do
-  types=$(az vm list-sizes --location $location --query="[].name" --output tsv 2>/dev/null || true)
+  types=$(az vm list-sizes --location $location --query="[].name" --output tsv 2>/dev/null | sort -u)
```
matt-boris, line 22: The `|| true` was there because there are a handful of Azure locations that this API doesn't work on, such as `*staging` named locations. I wanted to just continue through if any failures occurred.

I wonder if there's something we can pass at the end to `jq` that'll sort the json?

petemoore: strange, i didn't hit that issue. I ran it without errors. :-/

matt-boris: Strange is right 😄 I'll approve and we can merge and fix it later if we see my issue then!

matt-boris (approved):

petemoore:
although....

```
@@ -1,45362 +1,45218 @@
```

petemoore:
Running one last time with the `|| true` added back in to see if extra lines appear...

matt-boris (approved):

PR #740, opened 2024-06-06
No need to keep the AWS proj-taskcluster Windows 2022 pools

petemoore:
This replaces `proj-taskcluster/gw(-ci)-windows-2022` pools with what we had in `proj-taskcluster/gw(-ci)-windows-2022-azure` pools and removes the `proj-taskcluster/gw(-ci)-windows-2022-azure` pools altogether.

As part of the migration effort from AWS to Azure, we shouldn't really need to maintain AWS and Azure versions anymore, and can just standardise on the Azure versions.


lotas (approved):
lgtm

PR #741, opened 2024-06-06
New taskcluster Windows GPU pool in Azure

petemoore:
This updates the images to the latest ones built, and adds a new Windows Server 2022 image set and proj-taskcluster worker pool to use it. The image set is in Azure, is essentially the same as the existing Windows 2022 image set, but additionally has an NVIDIA GPU driver installed. Note, the bootstrap script is shared between the Windows 2022 image sets, and at runtime the script determines if an NVIDIA GPU is present, and installs the driver if one is. This helps to avoid copy/paste between powershell scripts, so that we still only have a single powershell script for bootstrapping our Windows workers.

### DIFF
--- a/config/imagesets.yml
+++ b/config/imagesets.yml
@@ -152,13 +152,13 @@ generic-worker-win2022:
       us-east-2: ami-0ec83ef759ed8591a
   azure:
     images:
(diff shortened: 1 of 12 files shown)


lotas (approved):
Awesome, thanks!

PR #742, opened 2024-06-07
Keep logs from installing python and NVIDIA driver in case of problems

petemoore:
Hoping this helps identify the problem with the failing python installation in Azure, and proactively capturing output from nvidia driver installation, as this is the first time we are doing it.


lotas (approved):
awesome!

lotas (approved):

PR #746, opened 2024-06-13
Move tutorial worker pool to aws

petemoore:
Needed for testing https://github.com/taskcluster/taskcluster/issues/7059


lotas (approved):

petemoore:
In hindsight it may be better to test kvm in the new pool `proj-taskcluster/gw-ubuntu-24-04-aws` that I created in #747, so I'll close this.

PR #747, opened 2024-06-13
Create ubuntu 24.04 image set and proj-taskcluster worker pools

petemoore:
Hoping to move all our workloads from Ubuntu 22.04 LTS to Ubuntu 24.04 LTS (and the update will upgrade podman to version 4, which is long overdue and hopefully will fix some open kvm issues).


lotas (approved):
:+1:

PR #749, opened 2024-06-14
Ensure IMDSv1 is enabled in AWS since worker-runner requires it

petemoore:
This helps avoid getting HTTP 401 when querying metadata service in EC2, if the base image that we build our AMIs from has IMDSv2 enforced.


lotas (approved):
wcpgw :)

PR #752, opened 2024-06-19
Staging Windows Server 2022 image set and worker pool for testing Windows fixes

petemoore:
This adds a staging Windows Server 2022 image set and worker pool that uses it, in order to test making changes to Windows Server 2022 bootstrap script, without needing to roll out to production.

We already had this for Ubuntu 22.04.


lotas (approved):
lgtm

PR #754, opened 2024-06-19
Fix black check to have non-zero exit code if formatting required

petemoore:
I spotted that `black` was reporting formatting issues but CI task was passing. This PR has two commits. The first one fixes the code format failure reporting (so commit should fail CI) and the second commit formats the code, and should pass.


lotas (approved):
👍

PR #758, opened 2024-06-20
Fix PATH issues on Windows (do not include Administrator folders)

### LINKED ISSUE #757: Task users on Windows have PATH which includes administrator user folders
@jschwartzentruber reported in [matrix](https://matrix.to/#/!whDRjjSmICCgrhFHsQ:mozilla.org/$FckOTwEJ_-7x6-im2rY14_9dGXTpwxc2pzcnzdOBJU0?via=mozilla.org&via=matrix.org&via=ungleich.ch) that an administrator user folder (`C:\Users\Administrator\AppData\Local\Microsoft\WindowsApps`) is included in the path of a task user.

This was due to a bug in the bootstrap script here: https://github.com/taskcluster/community-tc-config/blob/e07e1e6e5fd39597350d6669833194a455e30934/imagesets/generic-worker-win2022/bootstrap.ps1#L153

which is setting the Machine path (i.e. for all users) to include the `PATH` environment variable of the current (administrative) user.

Instead, this code should append entries to the current Machine `PATH` environment variable.

petemoore:
Partially implemented in staging already in PR #752 where a new staging pool was created for Windows. The only change in staging from the original bootstrap script was a change to [this line](https://github.com/taskcluster/community-tc-config/pull/752/files#diff-ef62ab221f06181bde907db8511e274bbdfdb7b1290e05801bd13d8a517fdcf0R153):

```diff
153c153
< [Environment]::SetEnvironmentVariable("PATH", $Env:Path + ";C:\Program Files\Vim\vim80;C:\go\bin;C:\Program Files\Git\cmd;C:\Program Files\nodejs;C:\Program Files\Python311", "Machine")
---
> [Environment]::SetEnvironmentVariable("PATH", [Environment]::GetEnvironmentVariable("PATH", "Machine") + ";C:\Program Files\Vim\vim80;C:\go\bin;C:\Program Files\Git\cmd;C:\Program Files\nodejs;C:\Program Files\Python311", "Machine")
```

The change was to set the Machine (global) `PATH` by modifying its current value, rather than by setting it based on the current value of the `PATH` environment variable (which is for the user running the script, i.e. an administrator user).

Comparing [original](https://community-tc.services.mozilla.com/tasks/XEULhk4ZQ32OTlbCepu5DA/runs/0/logs/public/logs/live.log#L49) vs [staging](https://community-tc.services.mozilla.com/tasks/F1mMqN6yTU-gQp5NQnExDQ/runs/0/logs/public/logs/live.log#L49) on Azure (not AWS where the bug was originally reported) the differences were:

* `C:\Windows\system32\config\systemprofile\AppData\Local\Microsoft\WindowsApps` was removed in staging task user `PATH` (as desired)
* `C:\Program Files\Git\cmd` and `C:\Program Files\nodejs\` were added in staging task user `PATH` (although they already existed in the PATH, so appeared twice - whoops)

This PR attempts to fix this in staging, and when confirmed to be working, to apply to production too.

Fixes #757

### DIFF
--- a/config/imagesets.yml
+++ b/config/imagesets.yml
@@ -196,13 +196,13 @@ generic-worker-win2022-staging:
   workerImplementation: generic-worker
   azure:
     images:
-      centralus: /subscriptions/8a205152-b25a-417f-a676-80465535a6c9/resourceGroups/rg-tc-eng-images/providers/Microsoft.Compute/images/imageset-oy0eijf5lgdpw3dr2vve-centralus
-      eastus: /subscriptions/8a205152-b25a-417f-a676-80465535a6c9/resourceGroups/rg-tc-eng-images/providers/Microsoft.Compute/images/imageset-oy0eijf5lgdpw3dr2vve-eastus
-      eastus2: /subscriptions/8a205152-b25a-417f-a676-80465535a6c9/resourceGroups/rg-tc-eng-images/providers/Microsoft.Compute/images/imageset-oy0eijf5lgdpw3dr2vve-eastus2
-      northcentralus: /subscriptions/8a205152-b25a-417f-a676-80465535a6c9/resourceGroups/rg-tc-eng-images/providers/Microsoft.Compute/images/imageset-oy0eijf5lgdpw3dr2vve-northcentralus
-      southcentralus: /subscriptions/8a205152-b25a-417f-a676-80465535a6c9/resourceGroups/rg-tc-eng-images/providers/Microsoft.Compute/images/imageset-oy0eijf5lgdpw3dr2vve-southcentralus
-      westus: /subscriptions/8a205152-b25a-417f-a676-80465535a6c9/resourceGroups/rg-tc-eng-images/providers/Microsoft.Compute/images/imageset-oy0eijf5lgdpw3dr2vve-westus
-      westus2: /subscriptions/8a205152-b25a-417f-a676-80465535a6c9/resourceGroups/rg-tc-eng-images/providers/Microsoft.Compute/images/imageset-oy0eijf5lgdpw3dr2vve-westus2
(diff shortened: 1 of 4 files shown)


petemoore:
Triggered [new task](https://community-tc.services.mozilla.com/tasks/SipLzapxR2qTZaK0M7-h-w/runs/0/logs/public/logs/live.log#L49) to compare to [original](https://community-tc.services.mozilla.com/tasks/XEULhk4ZQ32OTlbCepu5DA/runs/0/logs/public/logs/live.log#L49).

petemoore:
The diff is:

```diff
8,9c8,9
< C:\Windows\system32\config\systemprofile\AppData\Local\Microsoft\WindowsApps
< 
---
> C:\Program Files\Git\cmd
> C:\Program Files\nodejs\
12,13d11
< C:\Program Files\Git\cmd
< C:\Program Files\nodejs
16c14
< C:\Users\task_171887572939787\AppData\Local\Microsoft\WindowsApps
---
> C:\Users\task_171889575901975\AppData\Local\Microsoft\WindowsApps
```

which looks good! The order of the folders in the `PATH` is marginally different, but that should not be a problem as the impacted folders should not have identically named binaries inside of them.

`C:\Windows\system32\config\systemprofile\AppData\Local\Microsoft\WindowsApps` is no longer present, as desired (note, on AWS the path would be slightly different, due to the differences in user that runs the bootstrap script in Azure vs AWS).

The trailing slash in `C:\Program Files\nodejs\` should be immaterial.

The change of task user id from `task_171887572939787` to `task_171889575901975` is expected, as it is unique to the task.

Since everything worked as desired, I will push a last commit to apply the fixes to production image pool too.

lotas (approved):
awesome

PR #760, opened 2024-06-20
Switch back to 10 minute interval for mozCI decision task

Archaeopteryx:
Incident which required longer interval has been resolved.

### DIFF
--- a/config/projects/mozci.yml
+++ b/config/projects/mozci.yml
@@ -241,7 +241,7 @@ mozci:
       description: Run mozci classification tasks for new pushes
       owner: mcastelluccio@mozilla.com
       emailOnError: true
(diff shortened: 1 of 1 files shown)


config/projects/mozci.yml
```diff
@@ -241,7 +241,7 @@ mozci:
       description: Run mozci classification tasks for new pushes
       owner: mcastelluccio@mozilla.com
       emailOnError: true
-      schedule: ['*/60 * * * *'] # every hour
+      schedule: ['*/10 * * * *'] # every hour
```
lotas, line 244:
```suggestion
      schedule: ['*/10 * * * *']
```
we can remove that redundant and misleading comment also

lotas (approved):
Thanks!

PR #762, opened 2024-06-20
Restructure azure machine type offerings

petemoore:
This aligns the machine type offerings in Azure with the way we store this information for Amazon. It should hopefully make it easier to see from locations (regions) are affected when changes are detected.

Additionally this PR attempts to fix an issue whereby imageset.sh wouldn't run successfully if it could not launch a VM in an Azure location because the machine type needed for creating the worker image isn't available. It fixes it in the same way we do in Amazon, which is to skip a location (or region in aws terms) if the machine type (instance type in aws terms) is not available there.


generate/workers.py
```diff
@@ -455,24 +455,23 @@ def aws(
...
-def azure_machine_types_by_location():
+@lru_cache(maxsize=100)
+def azure_machine_types_in_location(location):
     """
     Return the set of machine types (such as "Standard_F32s_v2") in an Azure location.
 
       location: The Azure location, such as "centralus"
 
-    The instances are read from config/azure-machine-type-offerings.json and
+    The instances are read from JSON files in config/ec2-instance-type-offerings, and
```
lotas, line 465: this path is confusing?

```suggestion
    The instances are read from JSON files in config/azure-machine-type-offerings, and
```

lotas (approved):
awesome!

generate/workers.py
petemoore: eagle eye!!! amazing, thanks!

PR #769, opened 2024-06-25
Build Ubuntu staging worker pool from tc branch issue7076

petemoore:
This is to test IMDSv2 support in Worker Runner before landing the code fixes.

### DIFF
--- a/imagesets/generic-worker-ubuntu-22-04-staging/bootstrap.sh
+++ b/imagesets/generic-worker-ubuntu-22-04-staging/bootstrap.sh
@@ -6,7 +6,7 @@ exec &> /var/log/bootstrap.log
 ##############################################################################
 # TASKCLUSTER_REF can be a git commit SHA, a git branch name, or a git tag name
 # (i.e. for a taskcluster version number, prefix with 'v' to make it a git tag)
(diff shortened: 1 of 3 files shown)


lotas (approved):
WOohooo

imagesets/generic-worker-ubuntu-22-04-staging/bootstrap.sh
```diff
@@ -6,7 +6,7 @@ exec &> /var/log/bootstrap.log
 ##############################################################################
 # TASKCLUSTER_REF can be a git commit SHA, a git branch name, or a git tag name
 # (i.e. for a taskcluster version number, prefix with 'v' to make it a git tag)
-TASKCLUSTER_REF='main'
+TASKCLUSTER_REF='issue7076'
```
lotas, line 9:
```suggestion
TASKCLUSTER_REF='main'
```
?

petemoore: I've left it how it was, because otherwise the source code would be at odds with the built images. This means if people are interested to see what is deployed on staging, they can look at the file, and it will say where it was built from, rather than where we may want to build it from next time...

PR #788, opened 2024-08-08
Bump dependency versions to latest

petemoore:
I'm not sure if there is a more idiomatic way, I don't think we have the renovate bot set up for this repo.


lotas (approved):
I only know of pip freeze but this is also good imho

PR #805, opened 2024-09-03
ubuntu24.04 disable networkd-wait-online in google

lotas:
https://askubuntu.com/questions/1500473/enabling-networkmanager-makes-systemd-networkd-wait-online-fail
https://bugs.launchpad.net/ubuntu/jammy/+source/systemd/+bug/2036358

### DIFF
--- a/imagesets/generic-worker-ubuntu-24-04-staging/bootstrap.sh
+++ b/imagesets/generic-worker-ubuntu-24-04-staging/bootstrap.sh
@@ -134,6 +134,12 @@ systemctl enable worker
 
 retry apt-get install -y ubuntu-desktop ubuntu-gnome-desktop podman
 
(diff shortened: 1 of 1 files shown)


imagesets/generic-worker-ubuntu-24-04-staging/bootstrap.sh
```diff
@@ -134,6 +134,12 @@ systemctl enable worker
 
 retry apt-get install -y ubuntu-desktop ubuntu-gnome-desktop podman
 
+if [ '%MY_CLOUD%' == 'google' ]; then
```
lotas, line 137: since it was working fine on `aws` let's make it gcp only

petemoore (approved):
Nice work!

PR #806, opened 2024-09-03
disable unattended upgrades

lotas:


matt-boris (approved):
👍🏻

imagesets/generic-worker-ubuntu-24-04-staging/bootstrap.sh
```diff
@@ -166,6 +166,9 @@ if [ '%MY_CLOUD%' == 'aws' ]; then
   echo 'snd-aloop' >> /etc/modules
 fi
 
+# avoid unnecessary shutdowns during worker startups
+systemctl disable unattended-upgrades
+
```
petemoore, line 171: This is nice and simple. Let's test there are no issues reported in `/var/log/bootstrap.<whatever>` (i forget file path) in GCP and AWS to be sure this doesn't silently fail etc.

petemoore (approved):

PR #811, opened 2024-09-05
fix(ubuntu): wait for networking on 24.04 staging

matt-boris:
Taken from https://unix.stackexchange.com/a/126146.

### DIFF
--- a/imagesets/generic-worker-ubuntu-24-04-staging/bootstrap.sh
+++ b/imagesets/generic-worker-ubuntu-24-04-staging/bootstrap.sh
@@ -106,6 +106,9 @@ echo 127.0.1.1 taskcluster >> /etc/hosts
 cat > /lib/systemd/system/worker.service << EOF
 [Unit]
 Description=Start TC worker
+# start once networking is online
+Wants=network-online.target
+After=network-online.target
 
 [Service]
 Type=simple


imagesets/generic-worker-ubuntu-24-04-staging/bootstrap.sh
```diff
@@ -107,6 +107,10 @@ cat > /lib/systemd/system/worker.service << EOF
 [Unit]
 Description=Start TC worker
 
+# start once networking is online
+Wants=network-online.target
```
lotas, line 111: I wonder if this is enough or not since we disable this service down below 
maybe i has the same name for the NetworkManager one?

lotas (dismissed):
🤞

imagesets/generic-worker-ubuntu-24-04-staging/bootstrap.sh
petemoore: i would say let's test this manually before rebuilding images - what do you think?

petemoore (approved):
I wonder how we can best test this. Do we know how often it occurs in the other pool?

matt-boris:
Successful first task here https://community-tc.services.mozilla.com/tasks/LRbDRCWRQSa-eWwPRD4e-w/runs/0

PR #819, opened 2024-09-09
fix: don't delete old resources to make rollbacks easier

matt-boris:
Fixes https://github.com/taskcluster/community-tc-config/issues/607.

### DIFF
--- a/imagesets/imageset.sh
+++ b/imagesets/imageset.sh
@@ -18,7 +18,7 @@ function deploy {
   # Presumably bash and env must already be in the PATH to reach this point,
   # but let's keep them in the dependency list in case this list is
   # copy/pasted to any docs, etc. Having them here doesn't do any harm.
-  for command in aws az base64 basename bash cat chmod cut date dirname env find flock gcloud git head mktemp pass rm sed sleep sort tail touch which xargs yq; do
+  for command in aws az base64 basename bash cat chmod cut date dirname env find flock gcloud git head mktemp pass rm sed sleep sort tail touch tr which xargs yq; do
     if ! which "${command}" > /dev/null; then
       log "  \xE2\x9D\x8C ${command}"
       log "${0} requires ${command} to be installed and available in your PATH - please fix and rerun" >&2
@@ -193,7 +193,7 @@ function deploy {
   # Link to bootstrap script in worker type metadata, if generic-worker worker type
   if [ "$(yq r ../config/imagesets.yml "${IMAGE_SET}.workerImplementation")" == "generic-worker" ]; then
     BOOTSTRAP_SCRIPT="$(echo "${IMAGE_SET}"/bootstrap.*)"
-    yq w -i ../config/imagesets.yml "${IMAGE_SET}.workerConfig.genericWorker.config.workerTypeMetadata.machine-setup.script" "https://github.com/taskcluster/community-tc-config/blob/${IMAGE_SET_COMMIT_SHA}/imagesets/${BOOTSTRAP_SCRIPT}"
+    yq w -i ../config/imagesets.yml "${IMAGE_SET}.workerConfig.genericWorker.config.workerTypeMetadata.machine-setup.script" "https://raw.githubusercontent.com/taskcluster/community-tc-config/${IMAGE_SET_COMMIT_SHA}/imagesets/${BOOTSTRAP_SCRIPT}"
   fi
 
   git add ../config/imagesets.yml
@@ -229,11 +229,11 @@ function aws_find_old_objects {
   log "Querying old instances..."
   OLD_INSTANCES="$(aws --region "${REGION}" ec2 describe-instances --filters "Name=tag:ImageSet,Values=${IMAGE_SET}" --query 'Reservations[*].Instances[*].InstanceId' --output text)"
 
-  # find old amis
+  # find old snapshots
   log "Querying previous AMI..."
   OLD_SNAPSHOTS="$(aws --region "${REGION}" ec2 describe-images --owners self --filters "Name=name,Values=${IMAGE_SET} *" --query 'Images[*].BlockDeviceMappings[*].Ebs.SnapshotId' --output text)"
 
-  # find old snapshots
+  # find old amis
   log "Querying snapshot used in this previous AMI..."
   OLD_AMIS="$(aws --region "${REGION}" ec2 describe-images --owners self --filters "Name=name,Values=${IMAGE_SET} *" --query 'Images[*].ImageId' --output text)"
 }
@@ -285,8 +285,6 @@ function aws_update {
   AMI="$(echo $AMI_METADATA | sed 's/ .*//')"
   AMI_NAME="$(echo $AMI_METADATA | sed 's/.* //')"
   log "Base AMI is: ${AMI} ('${AMI_NAME}')"
-
-  aws_find_old_objects
 
   TEMP_SETUP_SCRIPT="$(mktemp -t ${UNIQUE_NAME}.XXXXXXXXXX)"
 
@@ -387,8 +385,6 @@ function aws_update {
     fi
     echo "AMI:         ${IMAGE_ID}"
   } > "aws.${REGION}.secrets"
-
-  aws_delete_found
 }
 
 ################## GOOGLE ##################
@@ -440,8 +436,6 @@ function google_update {
   # gcloud projects add-iam-policy-binding "${GCP_PROJECT}" --member serviceAccount:taskcluster-worker-manager@taskcluster-temp-workers.iam.gserviceaccount.com --role roles/compute.imageUser
 
   # Prefer no ssh keys, see: https://cloud.google.com/compute/docs/instances/adding-removing-ssh-keys
-
-  google_find_old_objects
 
   TEMP_SETUP_SCRIPT="$(mktemp -t ${UNIQUE_NAME}.XXXXXXXXXX)"
 
@@ -496,8 +490,6 @@ function google_update {
   done
 
   echo "projects/${GCP_PROJECT}/global/images/${UNIQUE_NAME}" > gcp.secrets
-
-  google_delete_found
 }
 
 ################## AZURE ##################
@@ -553,8 +545,6 @@ function azure_update {
     log "Cannot deploy in ${REGION} since machine type $(cat azure_base_instance_type) is not supported; skipping."
     return 0
   fi
-
-  azure_find_old_objects
 
   # Avoid using UNIQUE_NAME which may be too long, see e.g.
   # https://bugs.launchpad.net/ubuntu-kernel-tests/+bug/1779107/comments/2
@@ -701,8 +691,6 @@ function azure_update {
     echo "Password:  ${ADMIN_PASSWORD}"
     echo "Image:     ${IMAGE_ID}"
   } > "azure.${REGION}.secrets"
-
-  azure_delete_found
 }
 
 ################## Entry point ##################
--- a/imagesets/generic-worker-freebsd/bootstrap.sh
+++ b/imagesets/generic-worker-freebsd/bootstrap.sh
@@ -1,7 +1,7 @@
 #!/bin/csh
 
 # Version numbers ####################
-setenv TASKCLUSTER_VERSION v68.0.3
+setenv TASKCLUSTER_VERSION v68.0.4
 ######################################
 
 pkg update
--- a/imagesets/generic-worker-ubuntu-24-04/bootstrap.sh
+++ b/imagesets/generic-worker-ubuntu-24-04/bootstrap.sh
@@ -4,7 +4,7 @@ set -exv
 exec &> /var/log/bootstrap.log
 
 # Version numbers ####################
-TASKCLUSTER_VERSION='v68.0.3'
+TASKCLUSTER_VERSION='v68.0.4'
 ######################################
 
 function retry {
--- a/imagesets/generic-worker-win2022-staging/bootstrap.ps1
+++ b/imagesets/generic-worker-win2022-staging/bootstrap.ps1
@@ -1,4 +1,4 @@
-$TASKCLUSTER_VERSION = "v68.0.3"
+$TASKCLUSTER_VERSION = "v68.0.4"
 
 # use TLS 1.2 (see bug 1443595)
 [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
--- a/imagesets/generic-worker-win2022/bootstrap.ps1
+++ b/imagesets/generic-worker-win2022/bootstrap.ps1
@@ -1,4 +1,4 @@
-$TASKCLUSTER_VERSION = "v68.0.3"
+$TASKCLUSTER_VERSION = "v68.0.4"
 
 # use TLS 1.2 (see bug 1443595)
 [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12


imagesets/.taskcluster-version
```diff
@@ -0,0 +1 @@
+v68.0.4
```
lotas, line 1: nice!

petemoore:
I feel like if we pull the version out of the script so it appears as a parameter without an explicit value, it would be good to include the value that was used in the `workerTypeMetaData` object. At the moment I think we include a reference to the url of the script in github, but i think we could also include the token values that are substituted in.

e.g. currently we have something like:

```
[taskcluster 2024-09-05T12:56:24.222Z]     "machine-setup": {
[taskcluster 2024-09-05T12:56:24.222Z]       "maintainer": "pmoore@mozilla.com",
[taskcluster 2024-09-05T12:56:24.222Z]       "script": "https://github.com/taskcluster/community-tc-config/blob/6d1dc826958e5181f1066190eb6df95535b2c9f4/imagesets/generic-worker-win2022/bootstrap.ps1"
[taskcluster 2024-09-05T12:56:24.222Z]     },
```

but instead we could have something like:

```
[taskcluster 2024-09-05T12:56:24.222Z]     "machine-setup": {
[taskcluster 2024-09-05T12:56:24.222Z]       "maintainer": "pmoore@mozilla.com",
[taskcluster 2024-09-05T12:56:24.222Z]       "bootstrap": {
[taskcluster 2024-09-05T12:56:24.222Z]         "script": "https://raw.githubusercontent.com/taskcluster/community-tc-config/6d1dc826958e5181f1066190eb6df95535b2c9f4/imagesets/generic-worker-win2022/bootstrap.ps1",
[taskcluster 2024-09-05T12:56:24.222Z]         "substitutions": {
[taskcluster 2024-09-05T12:56:24.222Z]           "%TASKCLUSTER_VERSION%": "v68.0.3",
[taskcluster 2024-09-05T12:56:24.222Z]           "%MY_CLOUD%": "google"
[taskcluster 2024-09-05T12:56:24.222Z]         }
[taskcluster 2024-09-05T12:56:24.222Z]       }
[taskcluster 2024-09-05T12:56:24.222Z]     },
```

Alternatively, a copy of the script with the substitutions applied could be stored at some url somewhere, and we just link to the version with the substitutions applied. That may be the most ideal for the user, since it is the simplest, but the alternative above I also find quite reasonable.

matt-boris:
@petemoore I went ahead and added the `substitutions` section as well as now link to the raw version of the bootstrap script.

imagesets/imageset.sh
```diff
@@ -193,7 +193,10 @@ function deploy {
   # Link to bootstrap script in worker type metadata, if generic-worker worker type
   if [ "$(yq r ../config/imagesets.yml "${IMAGE_SET}.workerImplementation")" == "generic-worker" ]; then
     BOOTSTRAP_SCRIPT="$(echo "${IMAGE_SET}"/bootstrap.*)"
-    yq w -i ../config/imagesets.yml "${IMAGE_SET}.workerConfig.genericWorker.config.workerTypeMetadata.machine-setup.script" "https://github.com/taskcluster/community-tc-config/blob/${IMAGE_SET_COMMIT_SHA}/imagesets/${BOOTSTRAP_SCRIPT}"
+    TASKCLUSTER_VERSION="$(cat "${IMAGE_SET}"/.taskcluster-version)"
+    yq w -i ../config/imagesets.yml "${IMAGE_SET}.workerConfig.genericWorker.config.workerTypeMetadata.machine-setup.script" "https://raw.githubusercontent.com/taskcluster/community-tc-config/${IMAGE_SET_COMMIT_SHA}/imagesets/${BOOTSTRAP_SCRIPT}"
```
lotas, line 197: would it make sense to use json-e here? :) I guess there are probably no cli tooling

imagesets/imageset.sh
```diff
@@ -249,11 +252,13 @@ function aws_delete_found {
     log "No previous instances to terminate."
   fi
 
-  # deregister old AMIs
+  # deregister old AMIs, but keep the most recent version
```
lotas, line 255: Which part is actually keeping the most recent version below? I'm struggling to read bash below

imagesets/imageset.sh
```diff
@@ -362,7 +368,7 @@ function aws_update {
 
   log "Now snapshotting the instance to create an AMI..."
   # now capture the AMI
-  IMAGE_ID="$(aws --region "${REGION}" ec2 create-image --instance-id "${INSTANCE_ID}" --name "${IMAGE_SET} version ${IMAGE_SET_COMMIT_SHA} (${UUID})" --description "${IMAGE_SET} version ${IMAGE_SET_COMMIT_SHA} (${UUID})" --output text)"
+  IMAGE_ID="$(aws --region "${REGION}" ec2 create-image --instance-id "${INSTANCE_ID}" --name "${IMAGE_SET} version ${IMAGE_SET_COMMIT_SHA} (${UUID})" --description "${IMAGE_SET} version ${IMAGE_SET_COMMIT_SHA} (${UUID})" --tag-specifications "ResourceType=image,Tags=[{Key=taskcluster-version,Value=${TASKCLUSTER_VERSION}}]" --output text)"
```
lotas, line 371: nice

imagesets/imageset.sh
```diff
@@ -444,16 +452,18 @@ function google_update {
...
 
   TEMP_SETUP_SCRIPT="$(mktemp -t ${UNIQUE_NAME}.XXXXXXXXXX)"
+  # GCP doesn't allow periods in label values
+  TASKCLUSTER_VERSION="$(cat .taskcluster-version | tr '.' '_')"
 
   if [ -f "bootstrap.ps1" ]; then
     PLATFORM=windows
     echo '&{' >> "${TEMP_SETUP_SCRIPT}"
-    cat bootstrap.ps1 | sed 's/%MY_CLOUD%/google/g' >> "${TEMP_SETUP_SCRIPT}"
+    cat bootstrap.ps1 | sed 's/%MY_CLOUD%/google/g' | sed "s/%TASKCLUSTER_VERSION%/${TASKCLUSTER_VERSION}/g" >> "${TEMP_SETUP_SCRIPT}"
```
lotas, line 461: this will also set `XX_YY_ZZ` for the bootstrap as well, is it intended?

imagesets/imageset.sh
```diff
@@ -444,16 +452,18 @@ function google_update {
   google_find_old_objects
 
   TEMP_SETUP_SCRIPT="$(mktemp -t ${UNIQUE_NAME}.XXXXXXXXXX)"
+  # GCP doesn't allow periods in label values
+  TASKCLUSTER_VERSION="$(cat .taskcluster-version | tr '.' '_')"
```
lotas, line 456: probably better to have separate variable with those substitutions for labels only, but keep the original one for the bootstrap substitutions
```suggestion
  TASKCLUSTER_VERSION="$(cat .taskcluster-version)"
  TASKCLUSTER_VERSION_LABEL="$(cat .taskcluster-version | tr '.' '_')"
```

imagesets/imageset.sh
```diff
@@ -229,13 +232,13 @@ function aws_find_old_objects {
...
-  OLD_SNAPSHOTS="$(aws --region "${REGION}" ec2 describe-images --owners self --filters "Name=name,Values=${IMAGE_SET} *" --query 'Images[*].BlockDeviceMappings[*].Ebs.SnapshotId' --output text)"
-
   # find old snapshots
   log "Querying snapshot used in this previous AMI..."
-  OLD_AMIS="$(aws --region "${REGION}" ec2 describe-images --owners self --filters "Name=name,Values=${IMAGE_SET} *" --query 'Images[*].ImageId' --output text)"
+  OLD_SNAPSHOTS="$(aws --region "${REGION}" ec2 describe-images --owners self --filters "Name=name,Values=${IMAGE_SET} *" --query 'Images[*].BlockDeviceMappings[*].Ebs.SnapshotId' --output text)"
+
+  # find old amis
+  log "Querying previous AMI..."
+  OLD_AMIS="$(aws --region "${REGION}" ec2 describe-images --owners self --filters "Name=name,Values=${IMAGE_SET} *" --query 'Images[*].{ImageId: ImageId, TaskClusterVersion: Tags[?Key==`taskcluster-version`].Value|[0]}' --output text)"
```
petemoore, line 241: I think this only finds previous AMIs with the same version of taskcluster - will that mean we'll keep the older deployments around forever?

imagesets/imageset.sh
```diff
@@ -249,11 +252,13 @@ function aws_delete_found {
     log "No previous instances to terminate."
   fi
 
-  # deregister old AMIs
+  # deregister old AMIs, but keep the most recent version
   if [ -n "${OLD_AMIS}" ]; then
-    log "Deregistering the old AMI(s) ("${OLD_AMIS}")..."
+    SORTED_AMIS=$(echo "${OLD_AMIS}" | while read AMI_NAME VERSION; do echo "$AMI_NAME|$VERSION"; done | sort -t '|' -k 2 -V)
```
petemoore, line 257: TIL `sort -V`! Nice!

imagesets/imageset.sh
```diff
@@ -408,7 +414,7 @@ function google_find_old_objects {
   fi
 
   log "Querying previous images..."
-  OLD_IMAGES="$(gcloud compute images list --project="${GCP_PROJECT}" --filter="labels.image-set=${IMAGE_SET}" --format='table[no-heading](name)')"
+  OLD_IMAGES="$(gcloud compute images list --project="${GCP_PROJECT}" --filter="labels.image-set=${IMAGE_SET}" --format='value(name,labels.taskcluster-version)')"
```
petemoore, line 417: Same here, I guess this only captures old images with the same taskcluster version, so wouldn't clean up older deployments?

imagesets/imageset.sh
matt-boris: Ah great catch, thank you. Google's picky and doesn't allow periods in the tag values. I learned this, then went and changed the variable to use underscores instead, forgetting about this find/replace.

imagesets/imageset.sh
```diff
@@ -193,7 +193,10 @@ function deploy {
   # Link to bootstrap script in worker type metadata, if generic-worker worker type
   if [ "$(yq r ../config/imagesets.yml "${IMAGE_SET}.workerImplementation")" == "generic-worker" ]; then
     BOOTSTRAP_SCRIPT="$(echo "${IMAGE_SET}"/bootstrap.*)"
-    yq w -i ../config/imagesets.yml "${IMAGE_SET}.workerConfig.genericWorker.config.workerTypeMetadata.machine-setup.script" "https://github.com/taskcluster/community-tc-config/blob/${IMAGE_SET_COMMIT_SHA}/imagesets/${BOOTSTRAP_SCRIPT}"
+    TASKCLUSTER_VERSION="$(cat "${IMAGE_SET}"/.taskcluster-version)"
+    yq w -i ../config/imagesets.yml "${IMAGE_SET}.workerConfig.genericWorker.config.workerTypeMetadata.machine-setup.script" "https://raw.githubusercontent.com/taskcluster/community-tc-config/${IMAGE_SET_COMMIT_SHA}/imagesets/${BOOTSTRAP_SCRIPT}"
+    yq w -i ../config/imagesets.yml "${IMAGE_SET}.workerConfig.genericWorker.config.workerTypeMetadata.machine-setup.substitutions.%TASKCLUSTER_VERSION%" "${TASKCLUSTER_VERSION}"
+    yq w -i ../config/imagesets.yml "${IMAGE_SET}.workerConfig.genericWorker.config.workerTypeMetadata.machine-setup.substitutions.%MY_CLOUD%" "${CLOUD}"
```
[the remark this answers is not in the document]
matt-boris: Ok I'm just pulling this stuff out then in favor of my arguments here https://github.com/taskcluster/community-tc-config/pull/819#issuecomment-2350010785. If you think we should have this, I think it should be in a separate PR.

imagesets/imageset.sh
matt-boris: No, this is just looking for tags with _key_ equal to `taskcluster-version`. It lists them all. I'm not passing in the current version of taskcluster we're building the image for.

imagesets/imageset.sh
matt-boris: Same as for the aws comment, this is simply outputting the `taskcluster-version` label alongside of the image name. So output for this would look something like this:
```bash
image_name_123456789        v68_0_4
image_name_987654321        v61_0_9
```

imagesets/imageset.sh
matt-boris: Good question. Lemme give an example of how this works:
```bash
# example output:
#     image_name_123456789        v68.0.4
#     image_name_987654321        v61.0.9

SORTED_AMIS=$(echo "${OLD_AMIS}" | while read AMI_NAME VERSION; do echo "$AMI_NAME|$VERSION"; done | sort -t '|' -k 2 -V)
```
This reads in the ami name and version and then creates a string to sort from using a `|` character so the output would then look like this:
```bash
#     image_name_123456789|v68.0.4
#     image_name_987654321|v61.0.9
```
Now the `sort` command is instructed to sort based on the part of the string after the `|` character with `-V` being passed so it's sorting version strings so that v68.0.5 is not sorted as a higher version than v68.0.40, for example. Now we have:
```bash
#     image_name_987654321|v61.0.9
#     image_name_123456789|v68.0.4
```
Here's the next line:
```bash
AMIS_TO_DELETE=$(echo "${SORTED_AMIS}" | sed '$d' | cut -d '|' -f1)
```
The `sed '$d'` will remove the last line of output (which includes the image with the highest recent version). And then the `cut` command will leave us with:
```bash
#     image_name_987654321
```
which we can then pass into the cli to remove.

imagesets/imageset.sh
matt-boris: Yes, very much needed in this case because image two versions being compared `v68.0.5` and `v68.0.40`. The latter is the higher version, but a simple `sort` would not say so.

imagesets/imageset.sh
lotas: aah.. thanks! So the `'$d'` is doing the not so obvious (if you don't know) part. Got it.

matt-boris: Exactly. I was contemplating pulling this logic out into a python script to make it more apparent as to what's going on 😅

imagesets/imageset.sh
```diff
@@ -229,13 +229,13 @@ function aws_find_old_objects {
...
-  # find old amis
-  log "Querying previous AMI..."
-  OLD_SNAPSHOTS="$(aws --region "${REGION}" ec2 describe-images --owners self --filters "Name=name,Values=${IMAGE_SET} *" --query 'Images[*].BlockDeviceMappings[*].Ebs.SnapshotId' --output text)"
-
   # find old snapshots
   log "Querying snapshot used in this previous AMI..."
-  OLD_AMIS="$(aws --region "${REGION}" ec2 describe-images --owners self --filters "Name=name,Values=${IMAGE_SET} *" --query 'Images[*].ImageId' --output text)"
+  OLD_SNAPSHOTS="$(aws --region "${REGION}" ec2 describe-images --owners self --filters "Name=name,Values=${IMAGE_SET} *" --query 'Images[*].BlockDeviceMappings[*].Ebs.SnapshotId' --output text)"
+
+  # find old amis
```
petemoore, line 236: Is there a reason to swap the order from finding old amis and then old snapshots to the other way around? i don't mind, it just makes the diff a bit more complicated.

imagesets/imageset.sh
```diff
@@ -249,11 +249,12 @@ function aws_delete_found {
     log "No previous instances to terminate."
   fi
 
-  # deregister old AMIs
+  # deregister old AMIs, but keep the most recent version
   if [ -n "${OLD_AMIS}" ]; then
```
petemoore, line 253: If there is only one old ami, it will say it is deregistering old amis, and then not actually do that. I think better to just filter out the amis to keep directly when you grab them (an extra pipe when fetching old amis).
Alternatively you could move AMIS-TO_DELETE to outside this if block, and then this if block can check if AMIS_TO_DELETE is empty or not (rather than checking if OLD_AMIS is empty or not).

imagesets/imageset.sh
```diff
@@ -535,9 +537,12 @@ function azure_delete_found {
     log "No previous resource groups to delete."
   fi
 
+  # delete old images, but keep the most recent version
   if [ -n "${OLD_IMAGES}" ]; then
-    log "Deleting the old image(s) ("${OLD_IMAGES}")..."
-    az image delete --ids ${OLD_IMAGES} --no-wait true > /dev/null 2>&1
+    SORTED_IMAGES=$(echo "${OLD_IMAGES}" | while read IMAGE_NAME CREATION_TIME; do echo "$IMAGE_NAME|$CREATION_TIME"; done | sort -t '|' -k 2 -V)
+    IMAGES_TO_DELETE=$(echo "${SORTED_IMAGES}" | sed '$d' | cut -d '|' -f1 | tr '\n' ' ' | sed 's/ $//')
```
petemoore, line 543: No need to use `tr` here, there are already no quotes around `${IMAGES_TO_DELETE}` on line 545

imagesets/imageset.sh
```diff
@@ -18,7 +18,7 @@ function deploy {
   # Presumably bash and env must already be in the PATH to reach this point,
   # but let's keep them in the dependency list in case this list is
   # copy/pasted to any docs, etc. Having them here doesn't do any harm.
-  for command in aws az base64 basename bash cat chmod cut date dirname env find flock gcloud git head mktemp pass rm sed sleep sort tail touch which xargs yq; do
+  for command in aws az base64 basename bash cat chmod cut date dirname env find flock gcloud git head mktemp pass rm sed sleep sort tail touch tr which xargs yq; do
```
petemoore, line 21: `tr` isn't needed here, see comment below

imagesets/imageset.sh
matt-boris: I actually just swapped the comment lines above :) Looks like they got mixed up for some reason down the line.

imagesets/imageset.sh
matt-boris: I just kept the same format as the other cli calls so that all images are listed one after the next. Also, `tr` is already used elsewhere in the script, so adding `tr` to the list of executables should've already been done.

imagesets/imageset.sh
matt-boris: I should've added this beforehand when I added this https://github.com/taskcluster/community-tc-config/blob/main/imagesets/imageset.sh#L574

imagesets/imageset.sh
```diff
@@ -229,13 +229,13 @@ function aws_find_old_objects {
...
   OLD_INSTANCES="$(aws --region "${REGION}" ec2 describe-instances --filters "Name=tag:ImageSet,Values=${IMAGE_SET}" --query 'Reservations[*].Instances[*].InstanceId' --output text)"
 
-  # find old amis
-  log "Querying previous AMI..."
-  OLD_SNAPSHOTS="$(aws --region "${REGION}" ec2 describe-images --owners self --filters "Name=name,Values=${IMAGE_SET} *" --query 'Images[*].BlockDeviceMappings[*].Ebs.SnapshotId' --output text)"
-
   # find old snapshots
   log "Querying snapshot used in this previous AMI..."
-  OLD_AMIS="$(aws --region "${REGION}" ec2 describe-images --owners self --filters "Name=name,Values=${IMAGE_SET} *" --query 'Images[*].ImageId' --output text)"
+  OLD_SNAPSHOTS="$(aws --region "${REGION}" ec2 describe-images --owners self --filters "Name=name,Values=${IMAGE_SET} *" --query 'Images[*].BlockDeviceMappings[*].Ebs.SnapshotId' --output text)"
```
petemoore, line 234: I think you're still deleting the snapshot that backs the AMI that is now being kept. I think you need to keep both the AMI and the snapshot. It might be worth checking this. I believe the AMI points to a snapshot, so if you keep the AMI but not the snapshot, I think it won't work.

matt-boris: Ah crap, you're right Pete. Nice eye. I'm imagining I could just do a similar sort by creation time, do you think that'd be safe? Or should I specifically determine which snapshot is backing the ami?

petemoore (dismissed):
A great improvement!

imagesets/imageset.sh
```diff
@@ -219,58 +219,6 @@ function deploy {
 
 ################## AWS ##################
 
-function aws_delete {
-  aws_find_old_objects
-  aws_delete_found
-}
```
petemoore, line 225: I think this is still needed for the "delete" action - currently you can run `imageset.sh update` or `imageset.sh delete`...

petemoore (approved):
Nice!

PR #821, opened 2024-09-12
Add log ingestor hook to capture logs from failed fuzzing tasks

jschwartzentruber:
Is there any way to prevent recursion with a hook like this? eg. if this hook itself fails, that will trigger itself and so on. Maybe with a trigger schema?

### DIFF
--- a/config/projects/fuzzing.yml
+++ b/config/projects/fuzzing.yml
@@ -372,6 +372,19 @@ fuzzing:
           - secrets:get:project/fuzzing/fuzzmanagerconf
         routes:
           - notify.email.truber@mozilla.com.on-failed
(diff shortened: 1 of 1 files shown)


jschwartzentruber:
> <del>Is there any way to prevent recursion with a hook like this? eg. if this hook itself fails, that will trigger itself and so on. Maybe with a trigger schema?</del>

Second commit adds a trigger schema that would prevent the hook failure from triggering itself.

petemoore (approved):
Awesome! Let us know how it works out for you!

petemoore:
```
Creating Hook=project-fuzzing/failed-log-ingestor
```

PR #825, opened 2024-09-13
Migrate Dv5 VM sizes to Dsv3 in order to support ephemeral OS disks

petemoore:
See https://learn.microsoft.com/en-us/azure/virtual-machines/ephemeral-os-disks

Since Dv5 family doesn't support premium storage (and thus doesn't support ephemeral OS disks) we should move to Dsv3 which is closely matched, but supports premium storage.

This should prevent the many errors we are seeing here:
* https://community-tc.services.mozilla.com/worker-manager/proj-relman%2Fwin2022/errors
* https://community-tc.services.mozilla.com/worker-manager/proj-git-cinnabar%2Fwindows/errors

### DIFF
--- a/config/projects/git-cinnabar.yml
+++ b/config/projects/git-cinnabar.yml
@@ -25,8 +25,8 @@ git-cinnabar:
       vmSizes:
         Standard_F8s_v2: 1
         Standard_F16s_v2: 1
-        Standard_D8s_v5: 1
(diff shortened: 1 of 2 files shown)


lotas (approved):

petemoore:
```
Updating WorkerPool=proj-relman/win2022
Updating WorkerPool=proj-git-cinnabar/windows
```

PR #836, opened 2024-09-25
Log output from az commands if they fail, but not if they pass

petemoore:
Sometimes an `az` command fails silently. When it passes, we don't want to show the stdout and stderr, but if it fails we do. Previously we were just sending output to /dev/null but with this PR, we will write it to a temporary file, and then only log it if the command fails.


matt-boris (approved):
Neat, thanks!

PR #837, opened 2024-09-25
Delete old resource groups in azure to avoid hitting quotas

petemoore:
Hopefully this should fix the quota issues we have been hitting...

### DIFF
--- a/imagesets/imageset.sh
+++ b/imagesets/imageset.sh
@@ -508,17 +508,10 @@ function google_update {
 function azure_delete {
   azure_find_old_objects
   azure_delete_found
+  azure_delete_resource_groups
 }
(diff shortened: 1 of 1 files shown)


matt-boris (commented):
LGTM but, the delete should probably happen right after building so that we're not leaving resources around until the next time we build images.

petemoore:
I that other delete already exists, but if something crashes, they can stick around... See https://github.com/taskcluster/community-tc-config/blob/f157e3ff3d20069792cc79f6cad8ad1359b7e463/imagesets/imageset.sh#L694-L697

matt-boris (approved):
Thanks 👍🏻

PR #830, opened 2024-09-25
Install CUDA on Windows GPU workers

petemoore:
@marco-c does this look ok to you?

I don't think we currently have any Linux workers with GPUs - do you need that too?

Fixes #713

PR #840, opened 2024-09-26
Test reverting of CUDA installation on Windows staging

petemoore:
This reverts the CUDA changes on Windows Staging (which seems to have broken Windows GPU pools in #830).

Note, this reverts it only from staging, so that we don't test in production (even though production GPU pools are probably broken, it is a good habit to get into).

It might be worth trying one last time to build with the CUDA installation on staging, or alternatively, we could revert on staging, and then use a pool where we run as root, get an RDP session on it, running as root, and see what is going wrong when we manually RDP onto it and execute the installation steps.

Weirdly, the installation seems to happen, but it seems to break the worker somehow, because workers were getting spawned, but not calling Worker Manager to get credentials, as far as I can tell.

At the same time, I enabled interactive on all worker pools in community, because I can't think of a good reason we would ever want to disable it. If that changes, we can set it at a worker pool level later, but for now, I think we should just keep things simple and have it enabled everywhere, until that becomes a problem.

### DIFF
--- a/config/imagesets.yml
+++ b/config/imagesets.yml
@@ -76,7 +76,6 @@ generic-worker-freebsd:
     genericWorker:
       config:
         ed25519SigningKeyLocation: /etc/generic-worker/ed25519_key
-        enableInteractive: true
         idleTimeoutSecs: 15
         shutdownMachineOnIdle: true
         shutdownMachineOnInternalError: true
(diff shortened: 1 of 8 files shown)


lotas (approved):
🤞

PR #846, opened 2024-10-05
Big Windows staging bootstrap cleanup

petemoore:
This ended up being quite a few cleanups in one. I hope I didn't bite off too much.

I propose when we deploy this, we write a script that checks all of the installation steps, and then we run it as a task on the staging pool. If we can get it to work, we can make it part of the deployment steps, so we are sure that all the parts continue to work and don't silently fail, or succeed but not actually do what we intend them to do...

### DIFF
--- a/imagesets/generic-worker-win2022-staging/bootstrap.ps1
+++ b/imagesets/generic-worker-win2022-staging/bootstrap.ps1
@@ -3,8 +3,10 @@ $TASKCLUSTER_REF = "main"
 # use TLS 1.2 (see bug 1443595)
 [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
 
+md "C:\Install Logs"
+
 # capture env
-Get-ChildItem Env: | Out-File "C:\install_env.txt"
+Get-ChildItem Env: | Out-File "C:\Install Logs\env.txt"
 
 # needed for making http requests
 $client = New-Object system.net.WebClient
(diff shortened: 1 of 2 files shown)


lotas (approved):
great!

petemoore:
Testing in https://community-tc.services.mozilla.com/tasks/LDQ3SrrmQxSoejCLs4QhyQ

imagesets/generic-worker-win2022-staging/bootstrap.ps1
```diff
@@ -43,6 +49,22 @@ Stop-Service $servicesToDisable
 $servicesToDisable.WaitForStatus('Stopped', "00:01:00")
 $servicesToDisable | Set-Service -StartupType Disabled
 
+# skip OOBE (out of box experience)
```
jwmoss, line 52: I don't think this is needed for win2022, only 24H2 in my testing thus far.

petemoore: For community we are lucky that we don't have very complicated requirements, so whenever possible, we've tried to reduce the total number of scripts to as few as possible, and to share the same bootstrap scripts. So like Boris says, we just have a single powershell script that we use for both Windows Server 2022 and Windows 11. It may be that we just standardise on one of those in the end, depending on what workloads are needed in community, or we might keep them both if it is helpful for finding issues that are likely to hit fxci later on.

PR #847, opened 2024-10-07
Fix for errors in docker image building

petemoore:
See failures in e.g. https://github.com/taskcluster/taskcluster/commit/84fad78f45e09eee1edd3b984517faff00178d25


lotas (approved):
thanks for quick fix

petemoore:
I think we decided against this in the end? Do you guys remember?

petemoore:
I _think_ we decided to build using podman instead...

matt-boris:
I think this can be closed, as we just ended up removing `--init`, so we won't see this issue again (from me manually crafting the task to run without `--init` or using `podman`).

https://github.com/taskcluster/taskcluster/pull/7317

PR #848, opened 2024-10-07
Attempt to fix Windows Staging

petemoore:
Attempting to fix fallout from #846


petemoore:
only affects staging, so landing without review...

PR #849, opened 2024-10-08
Windows Staging cleanup

petemoore:
More cleanup for Windows Staging due to issues on Windows 11.


petemoore:
This only affects staging, so landing without review.

PR #851, opened 2024-10-08
This should enable us to build azure machine images in parallel

petemoore:
This should provide a significant performance boost when rolling out a new version of generic-worker (and friends) to all worker pools.

Note, I moved the deleting of the resource group outside of the locked region, since I suspect the deletions can occur in parallel. If this isn't the case, please let me know!


matt-boris (commented):
This is what I had at first, but was getting race condition failures like, cannot delete resource group while a move operation is in progress lol. I'm fine trying this again if you'd like to see if anything's changed!

matt-boris (approved):
Nice, thanks!

PR #853, opened 2024-10-08
imageset.sh: retry commands that can fail due to intermittent issues

petemoore:
Can't believe we didn't have this already, and hardly ever hit issues...


matt-boris (approved):
Nice, thanks 👍🏻

PR #854, opened 2024-10-08
Exit powershell on bootstrap failure

petemoore:
Before we land this, I just wanted to build one time without it, as a last chance to fix anything that might be broken (which should be visible from the logs in `C:\Install Logs`). Once there are no failures, it makes sense to enable this.


matt-boris (approved):

petemoore:
No longer needed (already applied in #860).

PR #855, opened 2024-10-08
Admin password fix

petemoore:
Ironically, the password is now slightly less secure (because there are fewer possible permutations), but still easily secure enough.


matt-boris (dismissed):
Nice, thank you!

matt-boris (approved):

PR #858, opened 2024-10-09
Performance optimisation - only blow away old resource groups in build-gw-images.sh

petemoore:
See https://github.com/petemoore/myscrapbook/blob/d47a141c6f3453dbe569ee5ce79a5a77208929de/build-gw-images.sh#L39-L42

This takes a lot of time during image building due to locking. This should save some time.

### DIFF
--- a/imagesets/imageset.sh
+++ b/imagesets/imageset.sh
@@ -708,9 +708,8 @@ function azure_update {
     --interval=15
 
   # Try to acquire an exclusive lock as only one `az resource move` can happen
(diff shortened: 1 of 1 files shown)


matt-boris (approved):
You think your build script should eventually be merged into this repo? That way a change like this would also coincide with https://github.com/petemoore/myscrapbook/blob/d47a141c6f3453dbe569ee5ce79a5a77208929de/build-gw-images.sh#L39-L42 within the same PR.

petemoore:
Absolutely it should. I'll do this in a new PR.

PR #861, opened 2024-10-11
Sign commits made to password store

petemoore:
Previously this was disabled since it required user interaction, however this was resolved with the setup steps described here:

  * https://github.com/taskcluster/community-tc-config/blob/39256b79ef66d378d3973c5a5d16f6dcc8544f98/imagesets/README.md?plain=1#L71-L91

Note, in addition, this should catch issues early when these steps haven't been performed or they were performed too long ago in the past, rather than the script failing at the end.


matt-boris (dismissed):
👍🏻

PR #863, opened 2024-10-11
`imageset.sh all`, to build and deploy all image sets in parallel

petemoore:
Follow on from #858


matt-boris (approved):
This is great! So, tackle the `TODOs` in a separate PR after this lands?

PR #868, opened 2024-10-11
Some tweaks to image set building

petemoore:
Hopefully this will:

  * reduce problems due to passphrases expiring
  * show reason why we got

```
imageset.sh: 2024-10-11T17:54:55Z: aws: generic-worker-win2022: us-east-1: Cannot deploy in us-east-1 since instance type c4.2xlarge is not supported; skipping.
imageset.sh: 2024-10-11T17:54:56Z: aws: generic-worker-win2022: us-east-2: Cannot deploy in us-east-2 since instance type c4.2xlarge is not supported; skipping.
imageset.sh: 2024-10-11T17:54:56Z: aws: generic-worker-win2022: us-west-1: Cannot deploy in us-west-1 since instance type c4.2xlarge is not supported; skipping.
imageset.sh: 2024-10-11T17:54:57Z: aws: generic-worker-win2022: us-west-2: Cannot deploy in us-west-2 since instance type c4.2xlarge is not supported; skipping.
```

since clearly it wasn't because the instance type wasn't supported (since it is). When we find out exactly what the error looks like when the instance type is not supported, we can grep it, and if the reason is something else, retry it and log it.


matt-boris (approved):
👍🏻

PR #871, opened 2024-10-14
chore: remove gpg preset passphrase setup

matt-boris:
I don't think these steps are necessary, so long as you have:
```
    default-cache-ttl 86400
    max-cache-ttl 86400
```
in your `~/.gnupg/gpg-agent.conf` file


petemoore (approved):
Thanks Boris! Let's give it a shot, it can't be worse! Sorry for all the pain...

PR #888, opened 2024-12-03
chore: small cleanup around installing kernel modules

matt-boris:
Cleanup from https://github.com/taskcluster/community-tc-config/pull/887. I hadn't realized we were installing `linux-modules-extra-$(uname -r)` which should be sufficient. I think the fix for GCP is creating the `/etc/modprobe.d/snd-aloop.conf` options file alongside adding `snd-aloop` to `/etc/modules`.


lotas (approved):

petemoore (approved):
Thanks Boris!

PR #889, opened 2024-12-06
Include new scope for pulse subscriptions into anonymous role

lotas:
New scope was introduced for graphql subscriptions to Pulse https://github.com/taskcluster/taskcluster/pull/7377


matt-boris (approved):
👍🏻

PR #891, opened 2024-12-09
feat: generate worker pool config with `workerManager` block too

matt-boris:
Fixes https://github.com/taskcluster/community-tc-config/issues/890.


lotas (approved):
Cool!

PR #892, opened 2024-12-09
fix: only generate `workerManager` config for azure pools

matt-boris:
Followup to https://github.com/taskcluster/community-tc-config/pull/891. We'll get 400's if we try to update worker pool configs with this additional field.


petemoore (approved):
LGTM, thanks Boris!

PR #896, opened 2024-12-18
fix: remove broken install link for gvim

matt-boris:
The recent windows image rollout for v77 failed due to the following issue in the install logs:
```powershell
PS>TerminatingError(Invoke-WebRequest): "<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>404 Not Found</title>
</head><body>
<h1>Not Found</h1>
<p>The requested URL was not found on this server.</p>
<hr>
<address>Apache/2.4.29 (Ubuntu) Server at artfiles.org Port 443</address>
</body></html>
"
```

Looking into the artfiles.org download for gvim, I see no more `.exe`s are available to download https://artfiles.org/vim.org/pc/. I think this is safe to remove so we're unblocked for now, as I believe it's only being installed for debugging purposes.


lotas (approved):
awesome

PR #900, opened 2025-01-09
chore: don't build staging images by default

matt-boris:
This will greatly reduce our time spent rebuilding production images.


petemoore (approved):
Many thanks Boris! Looks great. :-)

PR #904, opened 2025-01-21
Remove the covdiff fuzzing hook

pyoor:
The underlying data source has been broken for several months and it's not clear when or if it will be fixed.  Removing the hook for now.


matt-boris (approved):
Thanks 👍🏻

PR #917, opened 2025-02-12
chore: remove `runTasksAsCurrentUser` config and rename pools

matt-boris:
Due to the changes in https://github.com/taskcluster/taskcluster/pull/7505.

We'll also need a PR in the monorepo that points to these newly-named worker pools.


petemoore (approved):
I think this is the right thing to do for now, thanks for this Boris! We can follow up with a bit of a restructure once everything is working again.

PR #919, opened 2025-02-14
chore: stop building docker worker images during `./imageset.sh all`

matt-boris:
I don't think this is needed anymore, especially while building our production images for community tc. Seems like a waste of resources.


petemoore (approved):
Agree! Thanks Boris!

PR #920, opened 2025-02-19
chore: only build staging images in one region

matt-boris:
To save some resources...


petemoore (approved):
Looks awesome!

Note, you never need to update REGIONS because that is a superset, and if an image isn't built in that region, it will automatically be filtered out in the later loops. So you can remove all the lines where you have `REGIONS=(.....)`. They don't really do any harm, but it does increase the amount of code, so I'm also happy with removing those lines.

petemoore:
Actually, would we be able to do this via the imageset config rather than matching a substring of the imageset name? Image set names may change over time, and hardcoding logic based on name feels like it is something that could be hard to maintain, etc. Whereas if this is managed via config, it is easy to change what regions another image set is built for easily, which feels generally useful, rather than just special-casing staging. 😬  What do you think?

matt-boris:
Yeah that's fine by me, worst case would've been all regions would be built if the `-staging` suffix isn't matched. I can still switch though if you want?

PR #922, opened 2025-02-27
fix: adjust worker pool generation based on new launch config changes

matt-boris:
Calculates the launch config ID similar to how it's done in the DB for worker manager https://github.com/taskcluster/taskcluster/blob/3619a4113f056d13f1081d70eb30305da6018a5d/db/versions/0105-migration.sql#L25-L37 (aside from the fact that we're using `sha256` in this implementation).

Also moves `capacityPerInstance` to `workerManager` block in config, as previous config is deprecated.

Related to changes from https://github.com/taskcluster/taskcluster/pull/7266.

### DIFF
--- a/generate/workers.py
+++ b/generate/workers.py
@@ -150,7 +150,11 @@ class DynamicWorkerPoolSettings(WorkerPoolSettings):
         return True
 
     def supports_worker_manager_config(self):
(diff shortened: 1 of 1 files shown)


lotas (approved):
Thank you!!

PR #923, opened 2025-03-04
chore: bump `idleTimeoutSecs` for taskcluster CI pools for caches

matt-boris:
Should help make https://github.com/taskcluster/taskcluster/pull/7540 more beneficial to us.

### DIFF
--- a/config/projects/misc.yml
+++ b/config/projects/misc.yml
@@ -23,9 +23,6 @@ misc:
       cloud: gcp
       minCapacity: 0
       maxCapacity: 50
-      genericWorker:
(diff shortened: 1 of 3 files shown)


petemoore (approved):
This looks fine, it is just hard to know what impact it will have on cloud costs and cache hit ratio. It would be interesting to track total time (from push until last task resolved) with warm vs cold caches, and with longer and shorter idleTimeoutSecs, and then to plot this against cost, to understand where the sweet spot might be. Otherwise any changes we make may be shooting in the dark.

PR #925, opened 2025-03-05
Add rerun scope for project specific schedulers

jschwartzentruber:
Since we switched to `schedulerId: fuzzing`, I can't manually re-run fuzzing tasks anymore. I think this fixes it in a generic way.

### DIFF
--- a/config/grants.yml
+++ b/config/grants.yml
@@ -45,6 +45,9 @@
     - queue:create-task:highest:proj-<..>/*
     - queue:claim-work:proj-<..>/*
 
+    # project-specific schedulers
+    - queue:rerun-task:<..>/*
+
     # project-specific private artifacts
     - queue:get-artifact:project/<..>/*
 


matt-boris (approved):

matt-boris:
Applied:
```bash
Updating Role=project-admin:*
```

jschwartzentruber:
@matt-boris  Works now. Thank you!

petemoore:
It's a nice side-effect, but I suspect we should be setting the `projectId` property in the task definition, rather than `schedulerId`. 😬

The `schedulerId` property originates from former times when there was a separate Scheduler service to the Queue service, so its value today isn't very clear. However, given that we have `projectId` I really feel that should be where the project is specified, rather than the `schedulerId`... Then you would add the project specific rerun scope:

```
queue:rerun-task-in-project:<..>
```

Does this also work for you?

PR #926, opened 2025-03-05
Use GCP C3D machine type for fuzzing/ci pool

jschwartzentruber:
C3D gives much better performance than N2 for a small price increase.


matt-boris:
Applied:
```bash
Updating WorkerPool=proj-fuzzing/ci
```

PR #928, opened 2025-03-19
fix(linux): also install most recent kernel modules

matt-boris:
_Should_ fix https://github.com/taskcluster/taskcluster/issues/7582.

This aligns with the recent staging changes that are tested in the comment below https://github.com/taskcluster/community-tc-config/commit/d2dc92f9da774b9686d6d5dead092091d5fb8fa4.

### DIFF
--- a/imagesets/generic-worker-ubuntu-24-04/bootstrap.sh
+++ b/imagesets/generic-worker-ubuntu-24-04/bootstrap.sh
@@ -129,7 +129,20 @@ systemctl enable worker
 # Installs the snd-aloop, v4l2loopback kernel modules
 # used for the audio/video devices, and vkms
 # required by Wayland
+#
+# Installs the extra kernel modules for the currently
+# running kernel version as well as the cloud-specific
+# meta-package in case we upgrade to a new kernel version
+# on reboot
 retry apt-get install -y linux-modules-extra-$(uname -r)
+case '%MY_CLOUD%' in
+  google)
+    retry apt-get install -y linux-modules-extra-gcp
+    ;;
+  aws)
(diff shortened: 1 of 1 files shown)


matt-boris:
Yep you can see we're installing an old version https://community-tc.services.mozilla.com/tasks/NucMjbLiRBOsRd53nIieHQ/runs/0/logs/public/logs/live.log#L17696 and then with my change we now also install the latest https://community-tc.services.mozilla.com/tasks/NucMjbLiRBOsRd53nIieHQ/runs/0/logs/public/logs/live.log#L17792

I got the current kernel version on the image that you guys are using https://community-tc.services.mozilla.com/tasks/W-grIKLSTUadKROWoxoTVQ/runs/0/logs/public/logs/live.log#L37

imagesets/generic-worker-ubuntu-24-04/bootstrap.sh
```diff
@@ -129,7 +129,20 @@ systemctl enable worker
 # Installs the snd-aloop, v4l2loopback kernel modules
 # used for the audio/video devices, and vkms
 # required by Wayland
+#
+# Installs the extra kernel modules for the currently
+# running kernel version as well as the cloud-specific
+# meta-package in case we upgrade to a new kernel version
+# on reboot
 retry apt-get install -y linux-modules-extra-$(uname -r)
```
petemoore, line 137: i wonder if this line is still needed, or if the meta package would effectively install it for the current kernel, and when the kernel is upgraded, would automatically install it for the updated kernel version? in any case, i think it does no harm. awesome work!

petemoore (approved):
Are we happy that enough testing has been done in staging? Looks great.

PR #934, opened 2025-04-02
fix(azure): remove security type to fix VM creation

matt-boris:
Fixes https://github.com/taskcluster/community-tc-config/issues/933.


matt-boris:
I confirmed locally with a test that removing this flag fixes the issue.

lotas (approved):

PR #944, opened 2025-05-20
Deprecate use of imageset.sh

petemoore:
It has been agreed that images should be built by Release SRE moving forward, so I am deprecating the tooling here.


lotas:
Is there another tooling in place to replace this? Looks scary to drop everything if we don't have anything to replace it with at the moment

petemoore:
Yes we should only land this once everything is working the new way. But the PR is prepared and ready to go when that is the case. I'll mark it as draft, to help avoid that we land it. Unfortunately "draft" makes it appear that it isn't ready to review, when it is. Really the state is "ready for review, but not ready to merge" but that isn't an option in the github interface. 😝

PR #964, opened 2025-10-30
Adds support for armDeployment in azure launch configs

lotas:
Although launch config accept much more, here we only expect `templateSpecId` to be present and some `parameters` Linked template is expected to support existing parameters and be compatible with the way we deployed so far.

Once this is tested and confirmed to be working, we might switch all azure pools to use same template and make it a default behavior

### DIFF
--- a/config/azure.yml
+++ b/config/azure.yml
@@ -1,15 +1,21 @@
 ---
-# This file describes Azure constants used to build worker pools
-# It supports one top level dictionary:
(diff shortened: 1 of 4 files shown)


matt-boris (dismissed):
👍🏻 Thanks

matt-boris (approved):

config/azure.yml
matt-boris: Could you please update the documentation at the top of this file with the new `armDeployment` object?

PR #966, opened 2025-11-04
Fix broken CI git clone urls

petemoore:
It looks like the github api switched from populating some fields with the github clone url with the github api url. This should fix the CI.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -8,18 +8,15 @@ tasks:
     repo:
       $if: 'tasks_for == "github-push"'
       then:
-        clone_url: ${event.repository.url}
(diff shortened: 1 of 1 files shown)


.taskcluster.yml
```diff
@@ -8,18 +8,18 @@ tasks:
     repo:
       $if: 'tasks_for == "github-push"'
       then:
-        clone_url: ${event.repository.url}
-        url: ${event.repository.url}
+        clone_url: ${event.repository.html_url}
```
lotas: I'm not sure if we have some kind of github api token or authentication token, afaik html urls would be subject to rate limiting without authentication

PR #990, opened 2026-06-23
fix(fuzzing): remove unused hooks

pyoor:


lotas (approved):

lotas:
tc-admin showed it was already applied, so I only merged the PR here

Issue #983, opened 2026-04-01
macmini-m4-127 offline — not accepting jobs


petemoore:
## Summary

`macmini-m4-127` in the `gw-ci-macos` worker type is offline and not accepting jobs.

## Evidence

- **Worker page** shows no recent tasks: https://community-tc.services.mozilla.com/provisioners/proj-taskcluster/worker-types/gw-ci-macos/workers/proj-taskcluster/macmini-m4-127?sortBy=started&sortDirection=desc
- **Worker type listing** doesn't even show the worker: https://community-tc.services.mozilla.com/provisioners/proj-taskcluster/worker-types/gw-ci-macos?sortBy=Last%20Active&sortDirection=desc

## Action needed

Investigate why the worker is offline and bring it back into service.

petemoore:
## Update: Stuck RSR preventing recovery

After rebooting the machine (with automatic updates disabled via `defaults write`), the same problem recurs — `Installer Progress` runs on every boot, preventing third-party LaunchDaemons from loading.

### Cause

A **Rapid Security Response (RSR) update** (`25D771280a` for macOS 26.3.1) is stuck in a retry loop. It was first attempted on 2026-03-17, failed with a boot policy error:

```
SUMacControllerErrorCommitStashFailed=7748
Failed to commit stash: MSU_ERR_BOOTPOLICY_FAILURE(43)
Failed to update BootPolicy for splat-only update
```

Every reboot triggers `Installer Progress` to retry this failed RSR, which puts the system into a restricted boot mode where third-party LaunchDaemons in `/Library/LaunchDaemons/` are not loaded by launchd.

### Current state

- The machine has an MDM profile ("Disable Automatic Updates" via SimpleMDM/Mozilla) but the RSR was already staged before the profile took effect
- `launchctl bootstrap system /Library/LaunchDaemons/com.mozilla.genericworker.plist` works as a manual workaround after each boot
- The worker cannot survive a reboot until the stuck RSR is cleared

### Needed

MDM/relops intervention to either:
1. Clear the pending RSR on this machine
2. Or push macOS 26.4 cleanly, which would supersede the stuck 26.3.1 RSR

### Workaround

After each reboot, manually run:
```
launchctl bootstrap system /Library/LaunchDaemons/com.mozilla.genericworker.plist
```

rcurranmoz:
Synchronized with @matt-boris on this just now. Lmk if we can help

petemoore:
## Update — looking into this further

[...]

PR #991, opened 2026-07-08
Workaround for issue #983

petemoore:
Until the underlying issue is fixed upstream by Apple, this should keep our workers alive...


lotas (approved):
🤞

petemoore:
Thanks Yarik! Deployed to both macs, let's see how they get on now!

PR #1001, opened 2026-09-02
Re-add run-task-as-current-user scopes on release pools for releases

Eijebong:
51fad690ee12913ac7330b5dc4d38aeafa688f31 removed wildcard scopes on `taskcluster:*` and didn't realize that it was the only provider for the `run-task-as-current-user` scope for the release pool.


lotas (approved):
thanks!

PR #1011, opened 2026-09-11
Retire the proj-taskcluster CI resources

petemoore:
These workloads all moved to firefox-ci so are no longer needed on community.
Note, the azure-attested-document hook is still there, tracked for migration in https://github.com/taskcluster/taskcluster/issues/9149

The smoke tests entities I suspect may still be needed (they probably should have been proj-misc instead of proj-taskcluster since they pertain to the validation of the community-tc deployment rather than taskcluster CI).

Everything else I got rid of.

Note, the imageset building still updates the two-worker mac pool for proj-taskcluster, which now run in fxci instead of community. I think that is ok for now until management of the mac pool has been handed over to RelOps. I'm setting up a meeting for next week to discuss that.


lotas (approved):
👋 taskcluster, see you in another deployment

PR #1015, opened 2026-09-16
fuzzing: remove bugmon-pernosco pools from community-tc.

jschwartzentruber:
We have a g-w replacement managed in fuzzing-tc-config which I will rename to bugmon-pernosco after this is applied.


matt-boris (approved):
Sweet ty! Everything working well?

matt-boris:
Applied 👍🏻 If you want the pool deleted sooner than the cron job (I think once a day?) cleans it up, lmk.

